PR #1, opened 2018-02-08
Add packages

chrishajas:
This includes the following packages:

- structmatcher
- testhelper
- operating
- gplog
- dbconn
- cluster

We'd like to push these relatively soon. If you don't get a chance to look at them now, feel free to give us feedback and we can make changes at any time.


chrishajas:
In terms of naming, we currently have the Operating package which has a system struct that we use to mock out system functions using function pointers. Therefore, if you want to get the current time, instead of doing `time.Now()`, you would do `operating.System.Now()`. Is there a better name for this?

gplog/gplog.go
```diff
@@ -0,0 +1,270 @@
...
+
+	logger = NewLogger(os.Stdout, os.Stderr, logFileHandle, logfile, LOGINFO, program)
+}
+
+func SetLogger(log *GpLogger) {
+	logger = log
+}
+
+// This function should only be used for testing purposes
+func GetLogger() *GpLogger {
```
chrishajas, line 104: I would somewhat rather just take this out. I'm not sure it provides a lot of value.

adam8157:
Great work!

I think you could just merge this, and we users contribute if there is an issue.

PR #3, opened 2018-02-27
Ensure output functions are thread-safe

kyeap-vmware:
This commit adds a global mutex that is locked when any output functions (Info,
Error, etc.) are called and unlocked at the end.  While these output functions
were already reasonably thread-safe, adding a mutex ensures that calling any
output functions concurrently in different goroutines does not trigger the race
detector in Go or Ginkgo.

The mutex is a global variable instead of a local GpLogger variable to ensure
that modifying the logger with SetLogger during a call to one of the output
functions does not have any unexpected behavior due to the new logger's mutex
state differing from that of the old logger.

Co-authored-by: Jamie McAtamney <jmcatamney@pivotal.io>
Co-authored-by: Kevin Yeap <kyeap@pivotal.io>

### DIFF
--- a/gplog/gplog.go
+++ b/gplog/gplog.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"log"
 	"os"
+	"sync"
 
 	"github.com/greenplum-db/gp-common-go-libs/operating"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


khuddlefish (approved):
@kyeap-pivotal Does this now work with the ginkgo flag that detects race conditions? I'm good with this, but I also really hope you're not using SetLogger in production code >.>

jmcatamney:
@khuddlefish Yes, we wanted to make the change to stop Ginkgo flagging some gp_upgrade tests, and running tests that Ginkgo previously complained about with this fix and the flag enabled works fine.

And no, we're not using SetLogger in production, the concern was more about SetLoggers in tests, like a BeforeEach firing a SetLogger in one test when another one is calling Info, or the like, and causing flaky failures.  Probably never going to happen, but we just wanted to be sure.

chrishajas (approved):

PR #5, opened 2018-03-21
Enable executing remote commands on a per-host basis

jmcatamney:
Previously, the remote shell execution functions in the cluster package could
only execute shell commands on a per-segment basis.  This commit adds the
capability to execute them on a per-host basis.  As with the per-segment
functions, it is possible to specify whether to include or exclude the master
in the list of hosts on which to execute the function.

To avoid copy-pasting or drastically modifying the generic functions for command
execution and error handling, the new GenerateSSHCommandMapForHosts function
still generates maps from ints (content IDs) to slices of strings (the commands
and their arguments) to be used with ExecuteClusterCommand; however, the error
messages for such function calls are appropriate for the per-host behavior and
don't mention segments at all to aid in debugging.

---

This PR introduces a breaking change for any utility calling GenerateAndExecuteCommand, so those utilities will have to be updated accordingly.  I've already made those changes locally for gpbackup and gprestore, and will be committing those after this PR is merged.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -9,6 +9,7 @@ import (
 	"bytes"
 	"fmt"
 	"os/exec"
+	"strings"
 
 	"github.com/greenplum-db/gp-common-go-libs/dbconn"
 	"github.com/greenplum-db/gp-common-go-libs/gplog"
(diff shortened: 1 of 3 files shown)


adam8157 (commented):
Our project doesn't use SSH commands, `COPY PROGRAM` instead, considering the SSH setup (user, password, exchanging keys) is a little complicated for the utility outside of clusters.

However yes, this looks good to me.

chrishajas (approved):
Looks good!

PR #6, opened 2018-03-23
Log command being run when a cluster command fails

chrishajas:
Previously, we logged the return value and output of the command, but
the actual bash command being run is helpful to reproduce the issue.

This changes the output from 
```
20180322:17:15:00 gpbackup:pivotal:melville:012578-[DEBUG]:-Unable to create segment data pipe on segment 0 on host localhost with error exit status 127: bash: mkfiffo: command not found
```
to
```
20180322:17:15:00 gpbackup:pivotal:melville:012578-[DEBUG]:-Unable to create segment data pipe on segment 0 on host localhost with error exit status 127: bash: mkfiffo: command not found
 Command was: ssh -o StrictHostKeyChecking=no user@localhost mkfiffo /Users/user/workspace/gpdb/gpAux/gpdemo/datadirs/dbfast1/demoDataDir0/gpbackup_0_20180322171500_pipe
```

Authored-by: Chris Hajas <chajas@pivotal.io>

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -205,7 +205,8 @@ func (cluster *Cluster) CheckClusterError(remoteOutput *RemoteOutput, finalErrMs
 			if remoteOutput.Scope != ON_HOSTS {
 				segMsg = fmt.Sprintf("on segment %d ", contentID)
 			}
-			gplog.Verbose("%s %son host %s with error %s: %s Command was: %s", messageFunc(contentID), segMsg, cluster.GetHostForContent(contentID), err, remoteOutput.Stderrs[contentID], remoteOutput.CmdStrs[contentID])
+			gplog.Verbose("%s %son host %s with error %s: %s", messageFunc(contentID), segMsg, cluster.GetHostForContent(contentID), err, remoteOutput.Stderrs[contentID])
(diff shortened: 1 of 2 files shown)


cluster/cluster.go
```diff
@@ -202,7 +205,7 @@ func (cluster *Cluster) CheckClusterError(remoteOutput *RemoteOutput, finalErrMs
 			if remoteOutput.Scope != ON_HOSTS {
 				segMsg = fmt.Sprintf("on segment %d ", contentID)
 			}
-			gplog.Verbose("%s %son host %s with error %s: %s", messageFunc(contentID), segMsg, cluster.GetHostForContent(contentID), err, remoteOutput.Stderrs[contentID])
+			gplog.Verbose("%s %son host %s with error %s: %s Command was: %s", messageFunc(contentID), segMsg, cluster.GetHostForContent(contentID), err, remoteOutput.Stderrs[contentID], remoteOutput.CmdStrs[contentID])
```
jmcatamney: Could you put the "Command was" part in a separate Verbose call, so it's not all in one big line that might wrap two or three times?

khuddlefish (approved):

PR #8, opened 2018-04-11
Allow Begin()ing a transaction on each connection

adam8157:
We have a feature which requests creating and inserting tables in
parallel with multiple connections.

This commit allows rolling back a single transaction if errors happen.


jmcatamney (approved):
Looks good to me.  Thanks for taking care of that TODO for us!

chrishajas (approved):
Looks good!

PR #9, opened 2018-04-16
New a DB connection with database, username, host and port

adam8157:
New a DB connection with detailed arguments, so we could connect to
different clusters, the former one is name "NewDBConnFromEnvironment"
now.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -131,6 +131,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "a44e31fc3cf6e00873ef130ca4c23509a79426ba53c2632e074238d28925cc69"
+  inputs-digest = "9b48cdf280444383e8c5c9936d12ad65a8970fda7d9d77ac974c91c167d27307"
   solver-name = "gps-cdcl"
   solver-version = 1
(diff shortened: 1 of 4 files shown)


adam8157:
We have the feature request to connect multiple clusters, better to support it in common libs I suppose.

chrishajas:
Looks good. I'd also be in favor of instead changing NewDBConn to take 4 arguments (dbname, user, host port) and creating a new wrapper function (for example, NewDefaultDBConn) that only takes the dbname as an argument. That would require current applications to update their code, but it would also give us compile-time type checking.

cjcjameson:
+1 probably best to keep the argument types rather specific and have a
separate constructor.

jmcatamney:
+1 to having a type-checked constructor.  I'd name the wrapper constructor something more specific like NewDBConnFromEnvironment, as the actual logic in it is around environment variables and "DefaultDBConn" can imply a connection created with just the defaults of template1/gpadmin/mdw/5432.

adam8157:
Thanks for reviewing, type checking makes sense, will have another version.

adam8157:
Updated, please help reviewing.

I decided to pick a new name and keep the compatibility, and not having the default values in case accidentally connecting to other databases.

chrishajas (approved):
Logic looks good to me. I don't mind having to change function names if they're more explicit/better names since we version this dependency and can simply update the names once we bump the version. I agree with Jamie that calling the constructor "NewDBConnFromEnvironment" would be a more accurate name than "NewDBConn" or "NewDefaultDBConn".

adam8157:
I agree :) Update to the better names and merge.

PR #10, opened 2018-04-19
Switch lib/pq import to pf-qiu/pq for varbit type fix

pf-qiu:
lib/pq doesn't support bit and varbit type. We can't query length
of bit and varbit type via original library. We have created a PR
in upstream repo, hope they can merge it soon. In the mean time,
we switch to our custom fixed version. Then this commit can be
reverted after fix is merged into upstream.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -18,15 +18,6 @@
 
 [[projects]]
   branch = "master"
-  name = "github.com/lib/pq"
-  packages = [
-    ".",
-    "oid"
-  ]
-  revision = "88edab0803230a3898347e77b474f8c1820a1f20"
-
-[[projects]]
-  branch = "master"
   name = "github.com/onsi/ginkgo"
   packages = [
     ".",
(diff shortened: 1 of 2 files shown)


yydzero:
How about put pq under greenplum-db org, instead of personal repo?

adam8157:
https://github.com/golang/dep/blob/master/docs/Gopkg.toml.md#source

> A source rule can specify an alternate location from which the name'd project should be retrieved. It is primarily useful for temporarily specifying a fork for a repository.

chrishajas:
I'm apprehensive to include forks since they can become burdonsome to maintain and keep up to date, especially if the change doesn't get merged in upstream for some reason.

How are you planning to use this? If you're querying a table that has a bit or varbit type, would casting it to text such as `select varbit_col::text` be a workaround?

pf-qiu:
We have worked out a workaround to query catalog directly instead of rely on the wire protocol. Closing this PR.
If we encounter similar issues, I'm +1 on yydzero's approach, to create a forked version in greenplum organization.

PR #12, opened 2018-04-24
Add GetWithArgs and SelectWithArgs method to dbconn

pf-qiu:
Passing query argument is supported in lib/pq like this

q:= `SELECT a from t WHERE b = $1`
db.Query(q, arg1)

Add query argument support for Get and Select.
Only one varadic parameter is allowed for function, the whichConn
parameters is hardcoded to 0.

### DIFF
--- a/dbconn/dbconn.go
+++ b/dbconn/dbconn.go
@@ -244,11 +244,10 @@ func (dbconn *DBConn) MustExec(query string, whichConn ...int) {
 }
 
 func (dbconn *DBConn) GetWithArgs(destination interface{}, query string, args ...interface{}) error {
-	connNum := 0
-	if dbconn.Tx[connNum] != nil {
-		return dbconn.Tx[connNum].Get(destination, query, args...)
+	if dbconn.Tx[0] != nil {
+		return dbconn.Tx[0].Get(destination, query, args...)
 	}
-	return dbconn.ConnPool[connNum].Get(destination, query, args...)
+	return dbconn.ConnPool[0].Get(destination, query, args...)
 }
 
 func (dbconn *DBConn) Get(destination interface{}, query string, whichConn ...int) error {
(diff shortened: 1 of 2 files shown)


chrishajas:
This looks good to me. 

One concern: Having the connection as a variadic in some functions and the args as a variadic in others might be a bit confusing, and not being able to use them both together may be an issue in the future, but I'm not sure of a better way to do it at the moment.

While we also could just use Sprintf to interpolate the query, passing arguments directly to Get() and Select() is more secure and prevents SQL injection for some use cases so I think this is a good addition.

dbconn/dbconn.go
```diff
@@ -243,6 +243,14 @@ func (dbconn *DBConn) MustExec(query string, whichConn ...int) {
 	gplog.FatalOnError(err)
 }
 
+func (dbconn *DBConn) GetWithArgs(destination interface{}, query string, args ...interface{}) error {
+	connNum := 0
```
khuddlefish: Declaring a variable for connNum feels strange when the only value you use is 0

dbconn/dbconn_test.go
```diff
@@ -204,6 +204,23 @@ var _ = Describe("dbconn/dbconn tests", func() {
...
+			two_col_single_row := sqlmock.NewRows([]string{"schemaname", "tablename"}).
+				AddRow("schema1", "table1")
+			mock.ExpectQuery("SELECT (.*)").WithArgs(arg1, arg2).WillReturnRows(two_col_single_row)
+
+			testRecord := struct {
+				Schemaname string
+				Tablename  string
+			}{}
+
+			err := connection.GetWithArgs(&testRecord, "SELECT schemaname, tablename FROM two_columns ORDER BY schemaname", arg1, arg2)
```
khuddlefish: This doesn't appear to actually use the args you pass in. It would be nice to see a test case that does.

PR #14, opened 2018-05-04
Support custom prefix in gplog

pf-qiu:
Use SetLogPrefixFunc() to set a custom prefix function. By default,
the prefix function is nil and gplog use default log prefix as
"timestamp header level", which is the same as before.

### DIFF
--- a/gplog/gplog.go
+++ b/gplog/gplog.go
@@ -72,14 +72,16 @@ const (
  *          cannot be reached.  This function will exit the program after printing
  *          the error message.
  */
+type LogPrefixFunc func(string) string
 
 type GpLogger struct {
-	logStdout   *log.Logger
-	logStderr   *log.Logger
-	logFile     *log.Logger
-	logFileName string
-	verbosity   int
-	header      string
+	logStdout     *log.Logger
+	logStderr     *log.Logger
+	logFile       *log.Logger
+	logFileName   string
+	verbosity     int
+	header        string
+	logPrefixFunc LogPrefixFunc
 }
 
 /*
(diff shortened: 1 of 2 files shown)


gplog/gplog.go
```diff
@@ -72,14 +72,16 @@ const (
  *          cannot be reached.  This function will exit the program after printing
  *          the error message.
  */
+type LogPrefixFunc func(string) string
```
jmcatamney, line 75: Since this is only used in one place in the struct definition, I don't think we need to declare a type for this.

gplog/gplog.go
```diff
@@ -120,31 +122,42 @@ func GetLogger() *GpLogger {
...
-		logFileName: logFileName,
-		verbosity:   verbosity,
-		header:      getHeader(program),
+		logStdout:     log.New(stdout, "", 0),
+		logStderr:     log.New(stderr, "", 0),
+		logFile:       log.New(logFile, "", 0),
+		logFileName:   logFileName,
+		verbosity:     verbosity,
+		header:        GetHeader(program),
+		logPrefixFunc: nil,
```
jmcatamney, line 131: Let's set this to `defaultLogPrefixFunc` instead of nil, so that `GetLogPrefix` can just return the current value without a check.

gplog/gplog.go
```diff
@@ -120,31 +122,42 @@ func GetLogger() *GpLogger {
...
 	header := fmt.Sprintf(headerFormatStr, program, user, host, pid, "%s")
 	return header
+}
 
+func SetLogPrefixFunc(logPrefixFunc func(string) string) {
+	logger.logPrefixFunc = logPrefixFunc
 }
 
-func GetLogPrefix(level string) string {
+func defaultLogPrefixFunc(level string) string {
```
jmcatamney, line 149: We might want to add a `ResetLogPrefix` function to set it back to `defaultLogPrefixFunc`.

gplog/gplog.go
pf-qiu: I think this way it's easier to understand. Also user of this library can use that type.

gplog/gplog.go
pf-qiu: The problem lies in the InitializeLogging test cases. reflect.DeepEqual only treats nil function as equal. Even if the same function are treated not equal.

gplog/gplog.go
pf-qiu: You can just set it to nil to restore the default logic.

gplog/gplog.go
jmcatamney: Hmm.  Looking around, there's no good way to compare functions in Go by default.  This is probably fine, then, and we might want to add some function-comparison logic to the `structmatcher` package so we can use that in future tests.

gplog/gplog.go
jmcatamney: I was thinking in terms of the other change, to set `logPrefixFunc` to the default function instead of nil.  Since we have to leave that as nil for now, this is fine as-is.

gplog/gplog.go
jmcatamney: I guess it can't hurt for library users to be able to be more explicit with their function typing.

jmcatamney (approved):

PR #15, opened 2018-05-10
Add iohelper package for file reading and writing

jmcatamney:
This PR adds a new "iohelper" package containing various helper functions that make use of the OpenFileRead and OpenFileWrite functions in the operating package, to simplify common file reading and writing operations.

---

PR notes:

1) I went with `iohelper` as the package name because "io" is a built-in package and we established the "[foo]helper" pattern with `testhelper`.  If anyone has objections/suggestions, I'm all ears.

2) This isn't in the `operating` package (which would keep everything involving OpenFileRead/OpenFileWrite in one place) because that would cause an import loop between `operating` and `gplog`.

3) This PR also includes a small tweak to GetSegmentConfiguration because I noticed it while writing the iohelper tests and it wasn't big enough to deserve its own PR.  The intention is to make GetSegmentConfiguration usable with gpupgrade so that it can pull in more common-libs stuff.

4) The contents of this package are largely drawn from `utils/io.go` in gpbackup.  Once this goes in, I'll be following it up shortly with a gpbackup PR to remove the extracted code from io.go and use iohelper.go instead.

### DIFF
--- a/iohelper/iohelper.go
+++ b/iohelper/iohelper.go
@@ -1,0 +1,112 @@
+package iohelper
+
+/*
+ * This file contains generic file/directory manipulation functions that use
+ * the function pointers and types from the operating package, for easier test
+ * mocking around file IO.
+ */
+
+import (
+	"bufio"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"github.com/greenplum-db/gp-common-go-libs/operating"
+	"github.com/pkg/errors"
+)
+
+/*
+ * The following six OpenFileFor... and MustOpenFileFor... functions abstract
+ * the most common cases for wanting to open files: when reading, open a read-
+ * only handle and ignore file permissions as long as it's readable, and when
+ * writing or appending, create the file if it doesn't exist with relatively
+ * standard 644 permissions and then open it.
+ *
+ * If more niche or complex scenarios are involved, the underlying OpenFileRead
+ * and OpenFileWrite functions should be used directly, to grant more fine-
+ * grained control over flags and permissions.
+ */
+
+func OpenFileForReading(filename string) (operating.ReadCloserAt, error) {
+	fileHandle, err := operating.System.OpenFileRead(filename, os.O_RDONLY, 0644)
+	if err != nil {
+		return nil, errors.Errorf("Unable to open file for reading: %s", err)
+	}
+	return fileHandle, nil
+}
+
+func MustOpenFileForReading(filename string) operating.ReadCloserAt {
+	fileHandle, err := OpenFileForReading(filename)
+	gplog.FatalOnError(err)
+	return fileHandle
+}
+
+func OpenFileForWriting(filename string) (io.WriteCloser, error) {
+	flags := os.O_CREATE | os.O_WRONLY
+	fileHandle, err := operating.System.OpenFileWrite(filename, flags, 0644)
+	if err != nil {
+		return nil, errors.Errorf("Unable to create or open file for writing: %s", err)
+	}
+	return fileHandle, nil
+}
+
+func MustOpenFileForWriting(filename string) io.WriteCloser {
+	fileHandle, err := OpenFileForWriting(filename)
+	gplog.FatalOnError(err)
+	return fileHandle
+}
+
+func OpenFileForAppending(filename string) (io.WriteCloser, error) {
+	flags := os.O_APPEND | os.O_CREATE | os.O_WRONLY
+	fileHandle, err := operating.System.OpenFileWrite(filename, flags, 0644)
+	if err != nil {
+		return nil, errors.Errorf("Unable to create or open file for appending: %s", err)
+	}
+	return fileHandle, nil
+}
+
+func MustOpenFileForAppending(filename string) io.WriteCloser {
+	fileHandle, err := OpenFileForAppending(filename)
+	gplog.FatalOnError(err)
+	return fileHandle
+}
+
+func FileExistsAndIsReadable(filename string) bool {
+	_, err := operating.System.Stat(filename)
+	if err == nil {
+		var fileHandle io.ReadCloser
+		fileHandle, err = OpenFileForReading(filename)
+		if fileHandle != nil {
+			fileHandle.Close()
+		}
+		if err == nil {
+			return true
+		}
+	}
+	return false
+}
+
+// Read all lines from a file, stripping whitespace and ignoring empty lines
+func ReadLinesFromFile(filename string) ([]string, error) {
+	fileHandle, err := OpenFileForReading(filename)
+	if err != nil {
+		return nil, err
+	}
+	contents := make([]string, 0)
+	scanner := bufio.NewScanner(fileHandle)
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+		if line != "" {
+			contents = append(contents, line)
+		}
+	}
+	return contents, nil
+}
+
+func MustReadLinesFromFile(filename string) []string {
+	contents, err := ReadLinesFromFile(filename)
+	gplog.FatalOnError(err)
+	return contents
+}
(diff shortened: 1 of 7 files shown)


iohelper/iohelper.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func MustOpenFileForReading(filename string) operating.ReadCloserAt {
+	fileHandle, err := OpenFileForReading(filename)
+	gplog.FatalOnError(err)
+	return fileHandle
+}
+
+func OpenFileForWriting(filename string, allowAppend ...bool) (io.WriteCloser, error) {
+	flags := os.O_CREATE | os.O_WRONLY
```
chrishajas: I'm not sure if wrapping OpenFileWrite gives us much value. It's abstracting away the flags and permissions, which can be different (and important) in many cases.

iohelper/iohelper.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/greenplum-db/gp-common-go-libs/operating"
+	"github.com/pkg/errors"
+)
+
+/*
+ * Generic file/directory manipulation functions that use the function pointers
+ * and types from the operating package, for easier test mocking around file IO.
+ */
+
+func OpenFileForReading(filename string) (operating.ReadCloserAt, error) {
```
khuddlefish: I'm not sure these functions OpenFileForReading and OpenFileForWriting provide much value over the direct OpenFileRead and OpenFileWrite functions we already have. I would vote not to add these.

iohelper/iohelper.go
khuddlefish: I would rather see these options where we call the open file function rather than having them hidden in a wrapper

iohelper/iohelper.go
```diff
@@ -0,0 +1,95 @@
...
+	return readLines(fileHandle), nil
+}
+
+func MustReadLinesFromFile(filename string) []string {
+	fileHandle := MustOpenFileForReading(filename)
+	return readLines(fileHandle)
+}
+
+// Read all lines from a file, stripping whitespace and ignoring empty lines
+func readLines(fileHandle operating.ReadCloserAt) []string {
```
khuddlefish: Do you see people using this from a ReadCloser, or would they usually just want to read from a filename? If the latter, then maybe we can combine them into one function.

iohelper/iohelper.go
```diff
@@ -0,0 +1,95 @@
...
+	return fileHandle, nil
+}
+
+func MustOpenFileForWriting(filename string, allowAppend ...bool) io.WriteCloser {
+	fileHandle, err := OpenFileForWriting(filename, allowAppend...)
+	gplog.FatalOnError(err)
+	return fileHandle
+}
+
+func FileExistsAndIsReadable(filename string) bool {
```
chrishajas, line 76: We should pass through the error, otherwise we don't know whether the file didn't exist or couldn't be read due to permissions.

iohelper/iohelper_test.go
```diff
@@ -0,0 +1,213 @@
...
+			operating.System.OpenFileWrite = operating.System.OpenFileWrite
+		})
+		Describe("OpenFileForReading", func() {
+			It("creates or opens the file for reading", func() {
+				operating.System.OpenFileRead = func(name string, flag int, perm os.FileMode) (operating.ReadCloserAt, error) { return os.Stdin, nil }
+				fileHandle, err := iohelper.OpenFileForReading("filename")
+				Expect(err).ToNot(HaveOccurred())
+				Expect(fileHandle).To(Equal(os.Stdin))
+			})
+			It("panics on error", func() {
```
khuddlefish: This doesn't panic, it returns an error

iohelper/iohelper_test.go
```diff
@@ -0,0 +1,213 @@
...
+				operating.System.OpenFileWrite = func(name string, flag int, perm os.FileMode) (io.WriteCloser, error) {
+					passedFlags = flag
+					return os.Stdout, nil
+				}
+				fileHandle, err := iohelper.OpenFileForWriting("filename", true)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(fileHandle).To(Equal(os.Stdout))
+				Expect(passedFlags).To(Equal(os.O_APPEND | os.O_CREATE | os.O_WRONLY))
+			})
+			It("panics on error", func() {
```
khuddlefish: This also doesn't panic :P

iohelper/iohelper.go
khuddlefish: I'm not sure we would be parsing the error where this is used. Seems like it might just be "If FileExistsAndIsReadable do x else do y"

jmcatamney: Correct, we don't care what the error is in this case, just that it's not currently usable.

iohelper/iohelper_test.go
jmcatamney: Copy-paste, my bad.

iohelper/iohelper.go
jmcatamney: I know there was a reason I pulled this out instead of calling `ReadLinesFromFile` directly , but I don't recall what it was, so it must not have been that compelling.  I'll fold it back in.

jmcatamney:
@khuddlefish @chrishajas I'm willing to remove `OpenFileForReading`/`OpenFileForWriting` and fold the logic into the `Must...` functions, but I wanted to present my reasoning first and you can let me know if you'd still rather get rid of them.

The idea with those two functions was that in other packages we've been providing pairs of `Foo()` and `MustFoo()` functions that have the same arguments, error messages, etc. as one another but one panics and one returns an error, and this continues that pattern.  I think having those functions is helpful even if they're fairly light wrappers.

Having standard functions standardizes things, and 99% of the time when we want to write to a new file we're going to want to open it if it exists or create it if it doesn't, with a standard set of permissions; sticking that into a wrapper means you don't have to look those flags and permissions up each time, and the underlying function is still available if you want to use a nonstandard set of flags and permissions.

Also, having both versions around means it's easy to flip between `Foo` and `MustFoo` for debugging the way you might when e.g. using `regexp.Compile` to get the logic right and then switching to `regexp.MustCompile` when you're pretty sure it should always work in production, or starting off with `MustCompile` to easily expose errors before going back and bubbling them up the call stack properly.

Thoughts?

khuddlefish:
I suppose I didn't say it that clearly, but my "these don't seem to provide much value" comment applies to both the regular and "Must" version of the functions. If we do end up keeping these new wrappers, I like the idea of having separate regular ones that return an error and a "Must" version that panics.

Having the flags and modes hidden in there just seems like it's more likely to introduce errors. It is also a one liner function either way, it's just saving you the trouble of passing a few more arguments. In this case, I think it is more useful to have the visibility of those arguments rather than saving a few seconds writing it.

chrishajas:
I'm more okay with OpenFileForReading, however OpenFileForWriting seems risky. If open file didn't have those options, I would be fine with this. However, since the behavior of OpenFile can differ quite dramatically depending on flags, it's prone to misuse. For example, without the append flag, opening a file will not append any data to it, but will still succeed. This is something I would want a developer to spend some time to understand, especially since it may work fine in a simple unit test and only manifest itself in more complex cases (for example, our logger uses the append flag. It would still work if we removed it, but would overwrite all previous history).

jmcatamney:
If the concern is mostly about the hidden behavior of the `append` argument in `OpenFileForWriting`, how about this:

1) I'll split it into two functions, `OpenFileForWriting` and `OpenFileForAppending`, so someone can't clobber a file by accidentally forgetting to pass in a `true`.

2) I'll add a comment block before those functions explaning that these functions are intended to help with the most common, "happy path" behavior, and if someone wants to do something more complex they can use `OpenFileRead` and `OpenFileWrite` directly.

@khuddlefish @chrishajas Any objections?

@pf-qiu Do you have any additional comments on the PR, or have Chris and Karen covered everything?

chrishajas:
@jmcatamney Sounds good to me.

PR #16, opened 2018-05-23
Add errcheck linter

chrishajas:
This linter ensures that we check the error on functions that return an
error. This will help prevent bugs where we assume that a function such
as Chmod or Open is successful, but forget to check that no error is
returned.

In cases where we do intentionally want to ignore the error, we will now
need to assign it to an empty identifier, such as _ = os.Chmod().

In the this repo, I did not find any additional error checks that we were missing (If we can't write an error or log to stdout, there's not anything we can do beside panic()). However, I think the benefit of catching instances where we forget to check a legitimate error outweigh the additional syntax. Opinions?

Authored-by: Chris Hajas <chajas@pivotal.io>

### DIFF
--- a/dbconn/dbconn.go
+++ b/dbconn/dbconn.go
@@ -136,7 +136,7 @@ func (dbconn *DBConn) Close() {
 	if dbconn.ConnPool != nil {
 		for _, conn := range dbconn.ConnPool {
 			if conn != nil {
-				conn.Close()
+				_ = conn.Close()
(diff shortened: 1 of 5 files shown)


jmcatamney:
Yeah, it's ugly, but it's useful.  I'm in favor.

jmcatamney (approved):

adam8157 (approved):

PR #18, opened 2018-06-05
Update cluster to include dbid and port

jmcatamney:
This commit adds extra fields to cluster.Cluster and updates the query in
GetSegmentConfiguration to retrieve that information.  It also modifies the
structure of Cluster and its accessor functions to make it easier to add more
fields in the future if they're needed by any utilities.

---

This PR is mostly prompted by a desire to refactor ClusterPair in gpupgrade to just be two cluster.Clusters, and gpupgrade needs dbid and port information.  I'm not aware of any changes that would need to be made to existing utilities due to this refactor.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -27,13 +27,14 @@ type GPDBExecutor struct{}
 
 type Cluster struct {
 	ContentIDs []int
-	SegDirMap  map[int]string
-	SegHostMap map[int]string
+	Segments   map[int]SegConfig
 	Executor
 }
 
 type SegConfig struct {
+	DbID      int
 	ContentID int
+	Port      int
 	Hostname  string
 	DataDir   string
 }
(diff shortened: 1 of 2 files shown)


cluster/cluster.go
```diff
@@ -231,8 +230,20 @@ func (cluster *Cluster) GetContentList() []int {
 	return cluster.ContentIDs
 }
 
+func (cluster *Cluster) GetDbidForContent(contentID int) int {
+	return cluster.Segments[contentID].DbID
```
khuddlefish, line 234: Does this make sense? We should have two DbIDs per content ID so you shouldn't be able to get a single dbid as a result for one content id.

khuddlefish:
Are we just storing primaries in this structure, or do we have mirrors too? If we have mirrors, mapping by content ID might be tricky.

jmcatamney:
@khuddlefish Yes, we've been storing just the primaries in Cluster (c.f. the query in `GetSegmentConfiguration`), otherwise storing just one DataDir and Hostname per content wouldn't have made sense because primaries and mirrors can be on different hosts.

There shouldn't be any utilities that both need information about mirrors and would use this struct to store them, since for instance a Go port of gpaddmirrors wouldn't be able to query the database for as-yet-nonexistent mirrors, so sticking with primaries only should be safe.

chrishajas (approved):
There are a couple places in gpbackup that use `SegDirMap`, but they shouldn't be too difficult to change.

PR #19, opened 2018-06-14
add activerecord from kanas repo

violet2016:
active record is used by GPCC team. 
1. GPCC use this go lib to connect to gpdb
2. It has mockable interface to make test much easier. 
3. It support kerberos
4. It gives both programming way to write sql like Select("c1").From("table1"), and also a traditional way.

It originally located at the private repo "greenplum-db/kanas", we decided to move it to this public repo in https://www.pivotaltracker.com/story/show/156839319

It need greenplun-db/gp-golang-libpq (a kerberos supported pq) as the connection driver to greenplum. Also, gssapi lib (github.com/apcera/gssapi) is needed too.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,10 +2,25 @@
 
 
 [[projects]]
+  branch = "master"
+  name = "github.com/apcera/gssapi"
+  packages = ["."]
+  revision = "5fb4217df13b8e6878046fe1e5c10e560e1b86dc"
+
+[[projects]]
   name = "github.com/blang/semver"
   packages = ["."]
   revision = "2ee87856327ba09384cabd113bc6b5d174e9ec0f"
   version = "v3.5.1"
+
+[[projects]]
+  branch = "master"
+  name = "github.com/greenplum-db/gp-golang-libpq"
+  packages = [
+    ".",
+    "oid"
+  ]
+  revision = "fdb3a7bbb5e884323bbc7d88d17360621624aeb2"
 
 [[projects]]
   branch = "master"
@@ -149,6 +164,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "0898a326dc9933990618d3a5497e725420c9131a9d718d4927f24f6f02cc3939"
+  inputs-digest = "63a3708db895869e70d6001f3576216d15d443354c841ce8acd971e1f058f74a"
   solver-name = "gps-cdcl"
   solver-version = 1
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ lint :
 		gometalinter --config=gometalinter.config -s vendor ./...
 
 unit :
-		ginkgo -r -randomizeSuites -randomizeAllSpecs cluster dbconn gplog iohelper structmatcher 2>&1
+		ginkgo -r -randomizeSuites -randomizeAllSpecs cluster dbconn gplog iohelper structmatcher activerecord 2>&1
 
 test : lint unit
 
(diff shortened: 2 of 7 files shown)


jmcatamney:
Unfortunately, I'm going to have to strongly -1 adding this to gp-common-go-libs (at least in its current form), for a few reasons:

1) We've been trying to add packages to gp-common-go-libs only when one team working on a Go project wants to use something that another team has written, or when two teams are attempting to solve a similar problem, so that everything in the shared packages is being used by multiple teams, instead of putting lots of things into the shared library by default like gppylib did.

    I don't see any other teams making use of ActiveRecord considering that there are popular external query builder packages out there (e.g. [Squirrel](https://github.com/Masterminds/squirrel), [sqrl](https://github.com/elgris/sqrl), and [goqu](https://github.com/doug-martin/goqu)), so adding it to gp-common-go-libs doesn't provide much of a benefit.

2) Assuming there is a need for a custom query builder package, the packages in gp-common-go-libs are standardized on using a dbconn.DBConn for accessing the database, which uses the `sqlx` package to select into structs.  This is both to allow improvements/fixes to `dbconn` to benefit all of the other shared packages and to ensure that all the packages access the database in the same way so that developers only need to learn one package, set up tests in one way, and so forth.

    A new package that uses the built-in db.DB instead of dbconn.DBConn or even sqlx.DB and retrieves bare rows as `interface{}`s instead of selecting them into structs with `sqlx` breaks that pattern in two ways; at the very least, embedding a dbconn.DBConn and implementing `GetRows` and similar functions in a `sqlx` way would make it more interoperable with the other shared packages.  Additionally, if we're going to fork `libpq` to add functionality, the best package to add it to would be `dbconn` so all packages that use a dbconn.DBConn have those capabilities.

3) This package has a lot of duplicate functions (several SQL execution functions that duplicate `dbconn.DBConn` functionality, and many that reimplement variations on adding a token to an ActiveRecord instead of calling helper functions) and most of them are untested or only minimally tested.

    The SQL execution functions definitely need a _lot_ of comprehensive testing, and the token addition functions can definitely be generalized into one or two functions that can be completely tested and a bunch of wrappers around them.

4) I have several smaller issues with things like function names, unnecessary comments, `GetRow` being inefficient, and the like, but points 1 - 3 would have to be addressed before I'd be okay with adding ActiveRecord to gp-common-go-libs and then we could get to the normal code review stuff.

violet2016:
@jmcatamney Thanks for your reveiw. In this case, I'll move the activerecord to another repo instead of this one and close the pr.

magi345:
\+ @yydzero 

For some historical reason active record and libraries including `kanas/pq` `violet2016/gssapi` had been closed source and used only in GPCC project for years.
The intention to open-source them and contribute them into a common lib is to make more teams can leverage it.

Compared with `dbconn`, `activerecord` already have full support of kerberos and spnego which maybe useful for `gp-common-go-libs` users in future.
Current situation is more teams is using `dbconn`, when they comes to need kerberos they will have to choose,
1. Contribute and make `dbconn` support kerberos,
   - GPCC team have put everything to https://github.com/greenplum-db/gp-golang-libpq, free for refer and discussion
2. Use `activerecord`
   - It will be big effort if team go far with `dbconn`
   - [gp-golang-libpq](https://github.com/greenplum-db/gp-golang-libpq) is also open sourced and ready for use by other projects.

chrishajas:
For a new utility, are there advantages to using activerecord vs https://github.com/Masterminds/squirrel? SQL builders are common in many languages (including Go), so it would be nice to avoid building our own if possible.

yydzero:
Not sure about feature set.  squirrel is built around 2014, and we build activerecord around that time also, so as long as activerecord matches our expectation and squirrel has no outstanding advantages over activerecord, would suggest team to use what they have used.

So suggest to use separated repo instead of putting it to common go lib.

PR #20, opened 2018-06-29
Add LocalOutput to TestExecutor

nadeemg:
We need to specify outputs of command run locally. This seems the minimal change to get us working.


chrishajas (approved):
LGTM

PR #21, opened 2018-07-12
Use pgx postgres driver instead of pq

khuddlefish:
pq had an issue with character encoding for Thai characters, but pgx does
not seem to have the same issue. Performance seems to be the same and
the amount of code we need to change is very minimal.

Authored-by: Karen Huddleston <khuddleston@pivotal.io>


soumyadeep2007 (approved):

chrishajas:
Let's add a comment about why we're using pgx instead of libpq.

jmcatamney (approved):
The change is fine with me, but +1 on adding a comment as to why we're using a nonstandard driver.

khuddlefish:
Merged

PR #25, opened 2018-07-18
Improve SelectString and SelectStringSlice

jmcatamney:
Previously, these two functions used a generic struct with a "String" field to
execute the provided query with sqlx.Select.  This required any queries passed
to them to be of the form "SELECT [field] AS string" so that the results would
match the struct, and any fields that could not be aliased that way (such as
selecting GUCs with "SHOW [name]") could not be selected with these functions.

This commit modifies the functions to execute the query with Queryx instead of
Select, then scans in values from the underlying sql.Rows into sql.NullStrings,
which removes the need to alias fields in queries and allows for selecting GUCs
and NULL values with these functions.

Authored-by: Jamie McAtamney <jmcatamney@pivotal.io>

### DIFF
--- a/dbconn/dbconn.go
+++ b/dbconn/dbconn.go
@@ -306,9 +306,15 @@ func EscapeConnectionParam(param string) string {
 }
 
 /*
- * This is a convenience function for Select() when we're selecting single string
- * that may be NULL or not exist.  We can't use Get() because that expects exactly
- * one string and will panic if no rows are returned, even if using a sql.NullString.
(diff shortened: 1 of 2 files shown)


chrishajas:
Will this break existing code that uses the `SELECT AS string` pattern?

jmcatamney:
@chrishajas Nope, now SelectString doesn't care what the column header is, so you can `SELECT AS` anything you want (including `string`) or leave the field un-aliased.

chrishajas (approved):

PR #29, opened 2018-08-17
Improve GPDB version handling in dbconn

jmcatamney:
Previously, after connecting to GPDB with DBConn.Connect, a separate call to
DBConn.Version.Initialize(DBConn) was required to retrieve and set the version.
The original reasons for requiring a separate step to get the version no longer
apply, and having to make an extra call every time is irritating, so this commit
modifies DBConn.Connect to retrieve and set the version automatically.

It also adds a few new helper functions:

- dbconn.NewDBVersion, for creating sample GPDBVersions in tests.
- testhelper.CreateMockDBConn, for when you want to create a connection but
  don't want to connect yet.
- testhelper.ExpectVersionQuery, for setting a database version for a function
  that will Connect to a mock database with a DBConn.

----------

This PR makes a breaking change by altering `Initialize` and no longer requiring it to be called directly.  I've already made branches with the necessary changes to `gpbackup`, `gpcopy`, and `gpupgrade`, to be PR'ed once this PR goes in.

### DIFF
--- a/dbconn/version.go
+++ b/dbconn/version.go
@@ -5,7 +5,6 @@ import (
 	"strings"
 
 	"github.com/blang/semver"
-	"github.com/greenplum-db/gp-common-go-libs/gplog"
 )
 
 type GPDBVersion struct {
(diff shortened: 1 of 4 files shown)


slari-pivotal (approved):
Looks good.

adam8157 (approved):

dbconn/version.go
```diff
@@ -5,25 +5,42 @@ import (
...
-	gplog.FatalOnError(err)
+/*
+ * This constructor is intended as a convenience function for testing and
+ * setting defaults; the dbconn.Connect function will automatically initialize
+ * the version of the database to which it is connecting.
+ *
+ * The versionStr argument here should be a semantic version in the form X.Y.Z,
+ * not a GPDB version string like the one returned by "SELECT version()".
+ */
+func NewVersion(versionStr string) GPDBVersion {
```
jchampio: May want to document that an improper version string will result in a panic from this function.

jchampio (approved):

chrishajas (approved):

dbconn/version.go
jmcatamney: Will do.

PR #33, opened 2019-03-02
Add function to run local commands pertaining to segments

kyeap-vmware:
We have a way to generate ssh commands to run commands on remote
hosts, we do not have a way to run local commands pertaining to hosts. A
good example of how this could be used is if we needed to copy a file
from master to all segments

Just realized I didn't look into possible resource issues. Is spinning up 100+ scp processes as bad as getting pinged by 100+ scp processes started by other segment hosts?

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -41,17 +41,27 @@ type SegConfig struct {
 
 /*
  * We pass values from this enum into GenerateAndExecuteCommand to define the
- * scope for remote command execution.
- * - ON_SEGMENTS: Execute on each segment, excluding the master.
+ * nature and scope of the command execution.
+ * - ON_SEGMENTS:            Execute on each segment, excluding the master.
  * - ON_SEGMENTS_AND_MASTER: Execute on each segment, including the master.
- * - ON_HOSTS: Execute on each host, excluding the master host.
- * - ON_HOSTS_AND_MASTER: Execute on each host, including the master host.
+ * - ON_HOSTS:               Execute on each host, excluding the master host.
+ * - ON_HOSTS_AND_MASTER:    Execute on each host, including the master host.
+ *
+ * - ON_MASTER_TO_SEGMENTS:            Execute commands on master about segments, excluding master.
+ * - ON_MASTER_TO_SEGMENTS_AND_MASTER: Execute commands on master about segments, including master.
+ * - ON_MASTER_TO_HOSTS:               Execute commands on master about hosts, excluding master.
+ * - ON_MASTER_TO_HOSTS_AND_MASTER:    Execute commands on master about hosts, including master.
  */
 const (
 	ON_SEGMENTS = iota
 	ON_SEGMENTS_AND_MASTER
 	ON_HOSTS
 	ON_HOSTS_AND_MASTER
+
+	ON_MASTER_TO_SEGMENTS
+	ON_MASTER_TO_SEGMENTS_AND_MASTER
+	ON_MASTER_TO_HOSTS
+	ON_MASTER_TO_HOSTS_AND_MASTER
 )
 
 type RemoteOutput struct {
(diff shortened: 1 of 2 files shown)


cluster/cluster.go
```diff
@@ -226,6 +226,50 @@ func LogFatalClusterError(errMessage string, scope int, numErrors int) {
 	gplog.Fatal(errors.Errorf("%s on %d %s. See %s for a complete list of errors.", errMessage, numErrors, segMsg, gplog.GetLogFilePath()), "")
 }
 
+/*
+ * ExecuteMasterClusterCommands is a generic wrapper function
+ * to simplify execution of local commands pertaining to segments.
+ * e.g. running scp locally many times to push files to all segments
+ */
+func (cluster *Cluster) ExecuteMasterClusterCommands(generateCommand func(int) string) *RemoteOutput {
```
larham: I do like the argument of the anonymous function, but please weigh that with reuse benefits: the similarity between this method and the existing ExecuteClusterCommand(): is there an opportunity for reuse?

khuddlefish:
I think this might have similar load problems to the other method. We had been considering doing somewhat of a batched approach where you only run `n` commands at once and only start new ones once those are finished. This would probably be a good candidate for that. You could use a similar method, but just use buffered channels set to a certain size to facilitate it. I'm not sure what number would be a good default.

cluster/cluster.go
kyeap-vmware: +1 to this. The async logic can be combined. I think i'll still take an anonymous function though.

chrishajas:
The only commands I can think of that a user would use this function for are scp/rsync and maybe ping? Are there others you can think of? If not, it may be better to make this more specific to scp.

Also, consider how you'll want to do scp to each host instead of each segment. In the other function, we've passed in the `ON_SEGMENTS` or `ON_HOSTS` parameter.

With regard to resource issues, I don't think there will be a difference between sending out 100 ssh processes and 100 scp processes. It's definitely possible that there may be a limit we hit, but I don't think that would be specific to scp, ssh may have the same issue. However, sending out 100 ssh processes that scp back to the master is twice the number of connections than running 100 scp processes on the master host.

kyeap-vmware:
The biggest reason currently making this a generic function is because of the information required to construct a segment specific command for our use case. 

We are trying to copy an oid file over to all segments. Constructing the destination path of the oid file for each segment requires FilePathInfo, which is in the package of backup_filepath in the gpbackup repo. There would be a potential for a dependency loop or awkwardness for common-go-lib to have a dependency on a package located in gpbackup if we attempted to make the funcion specific to scp and rsync. 
The current implementation of moving plugin config using rsync gets around this by passing in an anonymous function.

It looks like there is potential to pull the generic cluster functionality of fpInfo into common-go-lib, but would requires significant effort.

kyeap-vmware:
The second commit does a better job of reusing the async property of existing code as suggested by @larham . If we like the second commit, I will drop the first commit.

oliverralbertini:
I was waiting to get some feedback on this before making a PR, but seems like it has become more relevant now: https://github.com/greenplum-db/gp-common-go-libs/pull/35.

cluster/cluster_test.go
```diff
@@ -248,6 +248,25 @@ var _ = Describe("cluster/cluster tests", func() {
...
+		It("Returns a map of local commands for one segment, including master", func() {
+			testCluster := cluster.NewCluster([]cluster.SegConfig{masterSeg, remoteSegOne})
+			commandMap := testCluster.GenerateLocalCommandMapForSegments(true, func(id int) string {
+				return fmt.Sprintf("echo %d", id)
+			})
+			Expect(len(commandMap)).To(Equal(2))
+			Expect(commandMap[-1]).To(Equal([]string{"bash", "-c", "echo -1"}))
+			Expect(commandMap[1]).To(Equal([]string{"bash", "-c", "echo 1"}))
+		})
+		It("Returns a map of local commands for one segment, excluding master", func() {
```
larham, line 261: This test has the same description as the one above it.

kyeap-vmware: the top one includes master, and this one excludes master.

Returns a map of local commands for one segment, **including** master
Returns a map of local commands for one segment, **excluding** master

kyeap-vmware:
Updated as per discussions and removed outdated commit.

larham (approved):
LGTM

chrishajas (approved):

kyeap-vmware:
We have discussed with @oliverralbertini about his contending PR: https://github.com/greenplum-db/gp-common-go-libs/pull/35 and have decided to proceed with this PR in favor of keeping the current API intact.

PR #35, opened 2019-03-04
Create cluster copy command

oliverralbertini:
The UD team uses the `gp-common-go-libs` in the PXF command-line utility. We recently implemented a feature that required copying files to remote hosts, but we were lacking a "copy" command.

Here is an implementation of a copy command that attempts to reuse as much code as possible from the ssh command. It uses rsync, maybe this can be configurable by the user to use scp.

In the process, I also decided to try to tackle the issue of `contentID` not being meaningful when doing commands/copies involving **hosts**. This required a change of the API.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,5 @@ _testmain.go
 *.prof
 
 vendor
+
+.idea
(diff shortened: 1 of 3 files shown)


oliverralbertini:
Seems relevant after seeing the discussion around https://github.com/greenplum-db/gp-common-go-libs/pull/33.

larham:
After discussing with @oliverralbertini , we incorporated many of the ideas from this PR into a separate one that did not require a breaking API change: https://github.com/greenplum-db/gp-common-go-libs/pull/33 , so please consider closing this one, unless the additional features here are required on top of that merged commit.

PR #36, opened 2019-03-06
Update CheckClusterError to handle new scopes

tyacovone:
These changes preserve the error messages of the previously supported scopes.

A test for the existing scope `ON_SEGMENTS_AND_MASTER` was missing, so we've added this test in addition to the four new `ON_MASTER_***` scopes.

Authored-by: Kevin Yeap <kyeap@pivotal.io>
Authored-by: Trevor Yacovone <tyacovone@pivotal.io>

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -251,11 +251,22 @@ func (cluster *Cluster) CheckClusterError(remoteOutput *RemoteOutput, finalErrMs
 
 	for contentID, err := range remoteOutput.Errors {
 		if err != nil {
-			segMsg := ""
-			if remoteOutput.Scope != ON_HOSTS && remoteOutput.Scope != ON_HOSTS_AND_MASTER {
-				segMsg = fmt.Sprintf("on segment %d ", contentID)
+			var dest string
+			hostname := cluster.GetHostForContent(contentID)
+			s := remoteOutput.Scope
+			switch {
+			case s == ON_SEGMENTS || s == ON_SEGMENTS_AND_MASTER:
+				dest += fmt.Sprintf("on segment %d ", contentID)
+				dest += fmt.Sprintf("on host %s", hostname)
+			case s == ON_HOSTS || s == ON_HOSTS_AND_MASTER:
+				dest += fmt.Sprintf("on host %s", hostname)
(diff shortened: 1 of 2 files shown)


cluster/cluster.go
```diff
@@ -251,11 +251,22 @@ func (cluster *Cluster) CheckClusterError(remoteOutput *RemoteOutput, finalErrMs
...
+			var dest string
+			hostname := cluster.GetHostForContent(contentID)
+			s := remoteOutput.Scope
+			switch {
+			case s == ON_SEGMENTS || s == ON_SEGMENTS_AND_MASTER:
+				dest += fmt.Sprintf("on segment %d ", contentID)
+				dest += fmt.Sprintf("on host %s", hostname)
+			case s == ON_HOSTS || s == ON_HOSTS_AND_MASTER:
+				dest += fmt.Sprintf("on host %s", hostname)
+			case s == ON_MASTER_TO_SEGMENTS || s == ON_MASTER_TO_SEGMENTS_AND_MASTER:
```
larham, line 263: I feel like we could simplify messages in the "run on master" set of cases.

In the case of running a command on master, consider how the command itself, like "scp somefile user@dest:/tmp" already might indicate all the information necessary, along with a typical error message about network or credentials.

tyacovone: I agree that it might be excessive... we were striving to be as consistent with the existing patterns as possible. I personally don't take issue with giving _too much_ error output

jmcatamney: We can definitely simplify things in the hosts-from-master case.  For the segments-from-master case, though, verbosity is good since commands won't necessarily have any mention of content id (e.g. it might only reference the segment data dir, which won't necessarily follow the `gpseg[content]` convention) and a lot of commands just give something like `ssh: exit status 1` instead of a useful message.

tyacovone: w/ @kyeap-pivotal : 
We agree that the case specified in lines 266-268 creates redundancy in the error message, but not including the `ON_MASTER_TO_HOSTS***` in our case statement may create some confusion down the line (esp given the ambiguity of which command (not always scp) is being used). 
If you both feel strongly about avoiding this redundancy, we can make the change, but we're both of the opinion that it provides more consistency and readability than ignoring the case all together.

larham (approved):
LGTM, the output messages can remain a little verbose for the sake of consistency

Issue #37, opened 2019-03-22
gplog log level for file output


bxu-github:
I am a member of the gp-command-center team.
We am interested in the gplog ( https://github.com/greenplum-db/gp-common-go-libs/blob/master/gplog/gplog.go ) you provided in this repo, 
but I have a small problem before I can use it:
some project can print a lot of logs into file, but it will enable user to choose which level info they'd like to log to the logFile, 
For example, if user chose 'Warn', then he won’t see Logs under 'Warn' level, he can only see these logs:  
- 'Fatal'
- 'Error'
- 'Warn'

We regard this a common use case,  Could you pls provide this possibility?

jmcatamney:
The gplog package has a `SetVerbosity` function to specify the logger's log level after it has been initialized.  To cause the logger to only display log messages with a level of Warning, Error, and Fatal, as in your example, you would call `gplog.SetVerbosity(gplog.LOGERROR)`.

Does that address your use case?

bxu-github:
Thanks for your reply. 
We know this `SetVerbosity` func.
Our use case is if set log level to `Info`.  We do not want **log file**  contains `debug` level log. But now the log file contains `debug` level log actually.
We don’t know how to achieve this.

jmcatamney:
Ah, I see what you mean now.  We currently default the logger to more verbose messages in the log file than at the terminal so that if a utility fails the user won't need to re-run it to get useful verbose output, but it shouldn't hurt to allow setting them separately as long as the logger still defaults to `INFO` for terminal output and `DEBUG` for log files.

I'll look into the cleanest way to make the change, given that existing utilities depend upon the current behavior.  Look for a PR to go up in the next few days.

bxu-github:
It's very helpful. 
Thanks

jimmyyih:
Closing issue.  Already fixed by https://github.com/greenplum-db/gp-common-go-libs/commit/aca9c3dc9103695d2b991c80eca40c1aeb67c38a.

PR #39, opened 2019-04-12
Add the ability to set logfile verbosity in gplog

jmcatamney:
This change was requested in issue #37.

To enable utilities using gplog to modify the verbosity of messages going to the log file, this PR splits the `verbosity` field in Logger into `shellVerbosity` and `fileVerbosity`, and functions have been added and modified to take advantage of it.

However, it does _not_ modify `InitializeLogging` to enable specifying the logfile verbosity and makes logfile verbosity an optional argument to `NewLogger`, as the initial reasons for wanting the logfile to default to `LOGDEBUG` still apply (namely, that it is important to log messages even if the user wants to suppress terminal output so that re-running utilities on error for better log output is not necessary) and leaving the logfile verbosity at `LOGDEBUG` instead of simply setting both shell and logfile verbosity to the same level should be discouraged.

------

This PR does not introduce any API changes, as existing functions retain their names and only optional arguments are added.

The specific feedback I'm looking for is whether anyone (A) objects to including this functionality at all or (B) objects to requiring consumers of gplog to go out of their way to modify logfile verbosity, and would rather having shell and logfile verbosity treated equally (at the expense of an API change).

### DIFF
--- a/gplog/gplog.go
+++ b/gplog/gplog.go
@@ -43,6 +43,10 @@ const (
 	 * The following constants representing the current logging level, and are
 	 * cumulative (that is, setting the log level to Debug will print all Error-,
 	 * Info-, and Verbose-level messages in addition to Debug-level messages).
+	 *
+	 * Log levels for terminal output and logfile output are separate, and can be
+	 * set independently.  By default, a new logger will have a verbosity of INFO
+	 * for terminal output and DEBUG for logfile output.
 	 */
 	LOGERROR = iota
 	LOGINFO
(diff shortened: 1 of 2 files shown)


oliverralbertini (approved):
I agree that file verbosity should be maintained at `LOGDEBUG` and I don't think it's necessary to allow the user to set verbosity for shell and file using the same function. In my opinion, that is not worth a change to the API.

bxu-github:
LGTM

Issue #41, opened 2019-04-28
gplog Filename and rotate


dujialun:
Hi, I am a member of the gp-command-center team. We're using gplog in our project. But we have some new requirements.
1. Set log file suffix myself. Now the log file name is followed by "_YYYYMMDD", but we want to set it to a simple format, like 'server.log' which is same as our former log name. Can you provide an interface like SetLogFileSuffix.
2. If we start the server and init the log, the file name won't change unless we restart the server and recall the log init function. We want to an interface with can rotate the log file, so it can change the writing file name to a new one.
3. When we call gplog.Fatal, it will trigger panic which will output the golang stack. If we put the stack info into terminal output, it is not user-friendly. So can you change the behavior to just exit(1).

PR #43, opened 2019-05-06
Add two features in gplog

dujialun:
- A new function FatalWithoutPanic, same as Fatal but will not panic and will exit(1)
- Enable user to customize the log file name

According to issue https://github.com/greenplum-db/gp-common-go-libs/issues/41

### DIFF
--- a/gplog/gplog.go
+++ b/gplog/gplog.go
@@ -40,6 +40,7 @@ var (
 	 * A function which can customize log file name
 	 */
 	logFileNameFunc		LogFileNameFunc
+	exitFunc	   		ExitFunc
 )
 
 const (
@@ -83,6 +84,7 @@ const (
  */
 type LogPrefixFunc func(string) string
 type LogFileNameFunc func(string, string) string
+type ExitFunc func()
 
 type GpLogger struct {
 	logStdout      *log.Logger
@@ -115,6 +117,14 @@ func InitializeLogging(program string, logdir string) {
 
 	createLogDirectory(logdir)
 
+	logfile := GenerateLogFileName(program, logdir)
+	logFileHandle := openLogFile(logfile)
+
+	logger = NewLogger(os.Stdout, os.Stderr, logFileHandle, logfile, LOGINFO, program)
+	SetExitFunc(defaultExit)
+}
+
+func GenerateLogFileName(program, logdir string) string {
 	var logfile string
 	if logFileNameFunc != nil {
 		logfile = logFileNameFunc(program, logdir)
@@ -122,9 +132,7 @@ func InitializeLogging(program string, logdir string) {
 		timestamp := operating.System.Now().Format("20060102")
 		logfile = fmt.Sprintf("%s/%s_%s.log", logdir, program, timestamp)
 	}
-	logFileHandle := openLogFile(logfile)
-
-	logger = NewLogger(os.Stdout, os.Stderr, logFileHandle, logfile, LOGINFO, program)
+	return logfile
 }
 
 func SetLogger(log *GpLogger) {
@@ -171,6 +179,10 @@ func SetLogPrefixFunc(logPrefixFunc func(string) string) {
 
 func SetLogFileNameFunc(fileNameFunc func(string, string) string) {
 	logFileNameFunc = fileNameFunc
+}
+
+func SetExitFunc(pExitFunc func()) {
+	exitFunc = pExitFunc
 }
 
 func defaultLogPrefixFunc(level string) string {
@@ -309,7 +321,7 @@ func FatalWithoutPanic(s string, v ...interface{}) {
 	errorCode = 2
 	_ = logger.logFile.Output(1, message)
 	_ = logger.logStderr.Output(1, message)
-	os.Exit(1)
+	exitFunc()
 }
 
 type stackTracer interface {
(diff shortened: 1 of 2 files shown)


gplog/gplog.go
```diff
@@ -286,6 +301,16 @@ func FatalOnError(err error, output ...string) {
 	}
 }
 
+func FatalWithoutPanic(s string, v ...interface{}) {
+	logMutex.Lock()
+	defer logMutex.Unlock()
+	message := GetLogPrefix("CRITICAL") + fmt.Sprintf(s, v...)
+	errorCode = 1
```
magi345: errorCode should be 2? see comments
```
/*
	 * Error code key:
	 *   0: Completed successfully (default value)
	 *   1: Completed, but encountered a non-fatal error (set by logger.Error)
	 *   2: Did not complete, encountered a fatal error (set by logger.Fatal)
	 */
```

gplog/gplog.go
```diff
@@ -286,6 +301,16 @@ func FatalOnError(err error, output ...string) {
 	}
 }
 
+func FatalWithoutPanic(s string, v ...interface{}) {
```
magi345, line 317: I'm not sure it is proper to add this function.
In comments of "abort()", it says
```
 * Abort() is for handling critical errors.  It panic()s to unwind the call stack
 * assuming that the panic is caught by a recover() in the main utility.
```
Does it means we should wrap gplog.Fatal() by ourselves and recover the panic to prevent printing stack trace?

magi345: However, I would like to have this function in gplog library because it makes our codebase neat.

magi345: For this PR I recommend add tests for this function and more comments at the top of this file, right after the comments of Info, Verbose, Debug, Warn, Error, Fatal

dujialun: > I'm not sure it is proper to add this function.
> In comments of "abort()", it says
> 
> ```
>  * Abort() is for handling critical errors.  It panic()s to unwind the call stack
>  * assuming that the panic is caught by a recover() in the main utility.
> ```
> 
> Does it means we should wrap gplog.Fatal() by ourselves and recover the panic to prevent printing stack trace?

I need a Fatal function without call stack, because when user see the stack, he may think our program is not robust, like accessing a null pointer. But if we wrap another function and recover the panic, our code would be tedious.

dujialun: > For this PR I recommend add tests for this function and more comments at the top of this file, right after the comments of Info, Verbose, Debug, Warn, Error, Fatal

It is impossible to add test case, for FatalWithoutPanic will exit, then the whole test progress will exit.

gplog/gplog.go
```diff
@@ -108,8 +114,14 @@ func InitializeLogging(program string, logdir string) {
 	}
 
 	createLogDirectory(logdir)
-	timestamp := operating.System.Now().Format("20060102")
-	logfile := fmt.Sprintf("%s/%s_%s.log", logdir, program, timestamp)
+
+	var logfile string
```
jmcatamney, line 128: For consistency with how the logfile prefix is handled, it might make more sense to pull this out into a `GenerateLogFileName` function (putting the existing logic in a `defaultLogNameFunc` function), so that it can be called both here and in any log rotation functions added later.

gplog/gplog.go
jmcatamney: > It is impossible to add test case, for FatalWithoutPanic will exit, then the whole test progress will exit.

It actually is possible to test `os.Exit` scenarios.  The recommended way is to use `exec.Command` to run the test in another process and check whether it exits (see [here](https://stackoverflow.com/questions/26225513/how-to-test-os-exit-scenarios-in-go) for an example of how to do this).

Another option would be adding a function pointer to Exit in the System Functions struct and mock it out with a function that just records that it was called instead of exiting the test run.

dujialun: > > It is impossible to add test case, for FatalWithoutPanic will exit, then the whole test progress will exit.
> 
> It actually is possible to test `os.Exit` scenarios. The recommended way is to use `exec.Command` to run the test in another process and check whether it exits (see [here](https://stackoverflow.com/questions/26225513/how-to-test-os-exit-scenarios-in-go) for an example of how to do this).
> 
> Another option would be adding a function pointer to Exit in the System Functions struct and mock it out with a function that just records that it was called instead of exiting the test run.

I mock the os.Exit by monkey.

gplog/gplog.go
dujialun: done, pls review

gplog/gplog_test.go
```diff
@@ -499,6 +500,17 @@ var _ = Describe("logger/log tests", func() {
 					gplog.Fatal(errors.New(expectedMessage), "")
 				})
 			})
+			FContext("FatalWithoutPanic", func() {
```
z-wenlin: Should remove F

dujialun:
@jmcatamney Can this PR be merged?

dujialun:
Per monkey's license:https://github.com/bouk/monkey/blob/master/LICENSE.md. We change the way mocking the exit function.

jmcatamney:
@dujialun Yep, looks good to me.

PR #44, opened 2019-05-08
Refactor command execution framework

jmcatamney:
This is a fairly extensive PR that makes large refactors to several aspects of Cluster and its command execution functions:
- Change per-host execution to not be a hack on top of per-segment execution
- Improve the way sets of commands and their outputs are stored
- Add support for mirror configuration in Cluster
- Simplify scope handling with bitmasking

See the individual commit messages for more details.

Note that this PR **does break the existing API** in several areas, but the actual changes required on the utility side should be fairly small (mostly changing scope variables passed into  `ExecuteClusterCommand` and tweaking generator functions to use hostnames instead of `GetHostForContent`) and by putting all the breaking commits into one PR they should hopefully be easier to deal with.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -19,15 +19,28 @@ import (
 
 type Executor interface {
 	ExecuteLocalCommand(commandStr string) (string, error)
-	ExecuteClusterCommand(scope int, commandMap map[int][]string) *RemoteOutput
+	ExecuteClusterCommand(scope int, commandList []ShellCommand) *RemoteOutput
 }
 
 // This type only exists to allow us to mock Execute[...]Command functions for testing
 type GPDBExecutor struct{}
 
+/*
+ * A Cluster object stores information about the cluster in three ways:
+ * - Segments is basically equivalent to gp_segment_configuration, a plain
+ *   list of segment information, and is ordered by content id.
+ * - ByContent is a map of content id to the single corresponding segment.
+ * - ByHost is a map of hostname to all of the segments on that host.
+ * The maps are only stored for efficient lookup; Segments is the "source of
+ * truth" for the cluster.  The maps actually hold pointers to the SegConfigs
+ * in Segments, so modifying Segments will modify the maps as well.
+ */
 type Cluster struct {
 	ContentIDs []int
-	Segments   map[int]SegConfig
+	Hostnames  []string
+	Segments   []SegConfig
+	ByContent  map[int]*SegConfig
+	ByHost     map[string][]*SegConfig
 	Executor
 }
 
@@ -40,8 +53,8 @@ type SegConfig struct {
 }
 
 /*
- * We pass values from this enum into GenerateAndExecuteCommand to define the
- * nature and scope of the command execution.
+ * We pass values from this enum into ShellCommands, RemoteOutputs, and associated
+ * functions to define the nature and scope of the command execution.
  * - ON_SEGMENTS:            Execute on each segment, excluding the master.
  * - ON_SEGMENTS_AND_MASTER: Execute on each segment, including the master.
  * - ON_HOSTS:               Execute on each host, excluding the master host.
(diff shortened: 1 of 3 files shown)


cluster/cluster.go
```diff
@@ -64,93 +77,162 @@ const (
...
-	}
-	return ConstructSSHCommand(cluster.GetHostForContent(contentID), cmdStr)
+/*
+ * A RemoteOutput is used to make it easier to identify the success or failure
+ * of a cluster command and to display the results to the user.
+ */
+type RemoteOutput struct {
+	Scope          int
+	NumErrors      int
+	Commands       []ShellCommand
```
kyeap-vmware, line 182: Maybe I'm missing something here, but the new implementation of a ShellCommand turns the storing the commands/outputs/errors from a map (`map[int]string`) to a slice (`[]Command`). This might cause compatibility issues because it removes the ability to retrieve specific output/cmd/err for a given segment based on contentID. With just a slice, in order to find an output of a specific Segment, a user would have to loop through all ShellCommands and ask what its contentID is. Verifying an output based on the contentID of a segment is something that might not be done a lot in production, but I could imagine tests doing so.

khuddlefish: You are right that it is more difficult to access specific command outputs per segment with the slice. My recollection of how gpbackup uses it and my guess/hope for how others would use it is that it mostly shouldn't matter when exact segment you are referencing because what matters most is if ANY segments encountered errors or if ANY segments had output, so you will loop through all of them anyways. It will change the format of test objects, but the slice might actually be easier to create than the map.

jmcatamney: Karen's correct, whenever you're running a command against the whole cluster you shouldn't care about individual segments, just the segments in aggregate.  The only case where you'd really want to get segment information one segment at a time would be to print them out in content order or the like, and moving from a map to a slice means that they're now ordered already.

jmcatamney:
@kyeap-pivotal @khuddlefish @chrishajas @oliverralbertini @frankgh 

Any further thoughts on this PR from anyone?  I'd like to try to get this merged soon if no one has any objections, since it's been sitting for a while.

oliverralbertini:
> Any further thoughts on this PR from anyone? I'd like to try to get this merged soon if no one has any objections, since it's been sitting for a while.

I haven't had time to give it a thorough review. I will do it by the end of this week.

cluster/cluster.go
```diff
@@ -64,93 +77,162 @@ const (
...
+ * - func(string) []string, which takes a hostname, for per-host commands
+ * The function uses a type switch to identify the right one, and panics if
+ * an invalid function type is passed in via programmer error.
+ * This method makes it easier for the user to pass in whichever function fits
+ * the kind of command they're generating, as opposed to having to pass in both
+ * content and hostname regardless of scope or using some sort of helper struct.
+ */
+func (cluster *Cluster) GenerateCommandList(scope int, generator interface{}) []ShellCommand {
+	commands := []ShellCommand{}
+	switch generateCommand := generator.(type) {
```
oliverralbertini, line 246: this is pretty sweet.

cluster/cluster.go
```diff
@@ -64,93 +77,162 @@ const (
...
+
+/*
+ * This function essentially wraps GenerateCommandList such that commands to be
+ * executed on other hosts are sent through SSH and local commands use Bash.
+ */
+func (cluster *Cluster) GenerateSSHCommandList(scope int, generator interface{}) []ShellCommand {
+	var commands []ShellCommand
+	localHost := cluster.GetHostForContent(-1)
+	switch generateCommand := generator.(type) {
+	case func(content int) string:
```
oliverralbertini, line 285: I think it might read nicer if we don't have the `content` or `host` when matching the interface type, since they will be used below again.

cluster/cluster.go
```diff
@@ -64,93 +77,162 @@ const (
...
+ * - func(int) []string, which takes a content id, for per-segment commands
+ * - func(string) []string, which takes a hostname, for per-host commands
+ * The function uses a type switch to identify the right one, and panics if
+ * an invalid function type is passed in via programmer error.
+ * This method makes it easier for the user to pass in whichever function fits
+ * the kind of command they're generating, as opposed to having to pass in both
+ * content and hostname regardless of scope or using some sort of helper struct.
+ */
+func (cluster *Cluster) GenerateCommandList(scope int, generator interface{}) []ShellCommand {
+	commands := []ShellCommand{}
```
oliverralbertini: as you have below, `var commands []ShellCommand` is recommended by my linter.

oliverralbertini: Same for `expectedCommands` in `cluster_test.go`

cluster/cluster_test.go
```diff
@@ -347,141 +233,85 @@ var _ = Describe("cluster/cluster tests", func() {
 		})
 		It("runs commands specified by command map", func() {
```
oliverralbertini: we should reword this to better reflect that `commandMap` -> `[]cluster.ShellCommand`

cluster/cluster_test.go
```diff
@@ -501,34 +331,55 @@ var _ = Describe("cluster/cluster tests", func() {
...
 			Expect(cluster.GetHostForContent(0)).To(Equal("localhost"))
-			Expect(cluster.Segments[3].DataDir).To(Equal("/data/gpseg3"))
+			Expect(cluster.Segments[2].DataDir).To(Equal("/data/gpseg3"))
 			Expect(cluster.GetHostForContent(3)).To(Equal("remotehost2"))
 		})
+		It("ensures that modifying a segment value in Segments is reflected in ByContent and ByHost", func() {
+			cluster := cluster.NewCluster([]cluster.SegConfig{masterSeg})
+			cluster.Segments[0].DataDir = "/new/dir"
+			Expect(cluster.GetDirForContent(-1)).To(Equal("/new/dir"))
+			Expect(cluster.GetDirsForHost("localhost")[0]).To(Equal("/new/dir"))
```
oliverralbertini: Why not also:
```go
Expect(len(cluster.GetDirsForHost("localhost"))).To(Equal(1))
```

cluster/cluster.go
oliverralbertini: For our use with PXF cluster utility, we don't need specific access to certain segments/hosts, and when we report errors, we just loop through the remote outputs, reporting errors as they come up.

I think this will work for us.

oliverralbertini (approved):
I made some minor comments regarding syntax mostly.

The ability to specifically select combinations of hosts vs segments, including vs excluding master or mirrors and to be able to run commands locally for file transfers makes this a much more complete API. And I like the bitmask approach as it will make adding future options easier, if the need arises.

This will be useful for the PXF cluster command utility where we have to sync files from master to all other hosts including segment hosts and standby master.

This looks great, thanks for doing this!

cluster/cluster.go
```diff
@@ -39,54 +39,106 @@ type Cluster struct {
...
- * - ON_MASTER_TO_HOSTS:               Execute commands on master about hosts, excluding master.
- * - ON_MASTER_TO_HOSTS_AND_MASTER:    Execute commands on master about hosts, including master.
+ * ON_SEGMENTS:     Execute one command per segment.
+ * ON_HOSTS:        Execute one command per host.
+ *
+ * INCLUDE_MASTER:  Include the master host or segment in the command list.
+ * EXCLUDE_MASTER:  Exclude the master host or segment from the command list.
+ *
+ * ON_REMOTE:       Execute each command on the specified remote segment/host.
+ * ON_LOCAL:        Execute all commands on the master host.
```
oliverralbertini, line 69: One thing that would be useful here (or in some documentation) is to let folks know that `ON_LOCAL` is the way to do file syncs/transfers. Of course it can be used for other things, like creating local files corresponding to each segment/host, etc.  But to someone new to using this API, it may not be obvious.

oliverralbertini: I just noticed the documentation of `GenerateAndExecuteCommand()`.  It is a good start, but it might be useful to clarify there that users can use the scope parameter, and in particular `cluster.ON_LOCAL` to achieve copy commands.

oliverralbertini: One final note about documentation: the comment at the top of the file should probably point folks straight to `GenerateAndExecuteCommand()` because with most of the functions being public, it can be confusing trying to figure out where to start using this library.

oliverralbertini (commented):
Just some documentation comments.

kyeap-vmware (approved):

PR #46, opened 2019-10-07
Add DBConn.ExecContext

jimmyyih:
Context allows queries to be cancelled. This can be used when we
expect a SIGINT/SIGTERM could occur during a query that could block
(e.g. trying to SELECT from a table but some other session explicitly
grabbed an AccessExclusiveLock and will not be giving it up anytime
soon). Without the ability to cancel queries, exiting Go programs
could lead to SQL session leaks. Simply closing the SQL connection
DOES NOT cancel the query for you.

### DIFF
--- a/dbconn/dbconn.go
+++ b/dbconn/dbconn.go
@@ -6,6 +6,7 @@ package dbconn
  */
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 	"strconv"
@@ -254,6 +255,19 @@ func (dbconn *DBConn) Exec(query string, whichConn ...int) (sql.Result, error) {
 
 func (dbconn *DBConn) MustExec(query string, whichConn ...int) {
 	_, err := dbconn.Exec(query, whichConn...)
+	gplog.FatalOnError(err)
+}
+
+func (dbconn *DBConn) ExecContext(queryContext context.Context, query string, whichConn ...int) (sql.Result, error) {
+	connNum := dbconn.ValidateConnNum(whichConn...)
+	if dbconn.Tx[connNum] != nil {
+		return dbconn.Tx[connNum].ExecContext(queryContext, query)
+	}
+	return dbconn.ConnPool[connNum].ExecContext(queryContext, query)
+}
+
+func (dbconn *DBConn) MustExecContext(queryContext context.Context, query string, whichConn ...int) {
+	_, err := dbconn.ExecContext(queryContext, query, whichConn...)
 	gplog.FatalOnError(err)
 }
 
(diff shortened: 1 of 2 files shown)


dbconn/dbconn.go
```diff
@@ -257,6 +258,19 @@ func (dbconn *DBConn) MustExec(query string, whichConn ...int) {
 	gplog.FatalOnError(err)
 }
 
+func (dbconn *DBConn) ExecContext(queryContext context.Context, query string, whichConn ...int) (sql.Result, error) {
```
lavjain, line 261: `ExecQueryWithContext` might be a more appropriate function name here.

jimmyyih: I'd rather not differ from the actual Golang database/sql ExecContext function name.  This dbconn wrapper seems to be intended for doing a bit of extra GPDB-related stuff before running the actual function.  I wouldn't want to remember two function names.  If so, we should rename a lot of the functions in this file to be GPDB-specific instead (e.g. Exec -> GPDBExecQuery, Select -> GPDBSelectQuery, etc) to avoid possible confusion.

jmcatamney: > This dbconn wrapper seems to be intended for doing a bit of extra GPDB-related stuff before running the actual function.

Yep, the intention is that dbconn functions wrap built-in function behavior, and someone using `database/sql` should in theory be able to switch to `dbconn` without renaming everything.

dbconn/dbconn_test.go
```diff
@@ -183,6 +184,36 @@ var _ = Describe("dbconn/dbconn tests", func() {
...
+			mock.ExpectExec("INSERT (.*)").WillReturnResult(fakeResult)
+			mock.ExpectCommit()
+
+			connection.MustBegin()
+			res, err := connection.ExecContext(ctx, "INSERT INTO pg_tables VALUES ('schema', 'table')")
+			connection.MustCommit()
+			Expect(err).ToNot(HaveOccurred())
+			rowsReturned, err := res.RowsAffected()
+			Expect(rowsReturned).To(Equal(int64(1)))
+		})
```
jmcatamney, line 215: The above tests don't test the behavior when you actually cancel the statement during execution.  You'll want to add some tests for that (at least one in and one outside a transaction) if you can find a non-hacky way to make that work with sqlmock.

jimmyyih: There is no non-hacky way to make an actual cancel test with sqlmock.  The actual implementation for canceling a query is half-baked at best (calling the cancel function doesn't give you feedback that the query has been canceled; you'd have to look at pg_stat_activity after canceling).

jmcatamney (approved):

shivzone (approved):

PR #47, opened 2019-11-07
structmatcher: Add a GomegaMatcher

dsharp-pivotal:
Could use some feedback on if `NegatedFailureMessage` is doing the right thing. Other than the negated failure message test, the added tests are copied from the existing tests.

### DIFF
--- a/structmatcher/structmatcher.go
+++ b/structmatcher/structmatcher.go
@@ -67,6 +67,7 @@ func StructMatcher(expected interface{}, actual interface{}, shouldFilter bool, 
 	return mismatches
 }
 
+// Deprecated: Use structmatcher.MatchStruct() GomegaMatcher
 func ExpectStructsToMatch(expected interface{}, actual interface{}) {
 	mismatches := StructMatcher(expected, actual, false, false)
 	if len(mismatches) > 0 {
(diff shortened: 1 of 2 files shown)


structmatcher/structmatcher.go
```diff
@@ -87,3 +87,48 @@ func ExpectStructsToMatchIncluding(expected interface{}, actual interface{}, inc
...
+type Matcher struct {
+	expected        interface{}
+	includingFields []string
+	excludingFields []string
+	mismatches      []string
+}
+
+var _ types.GomegaMatcher = &Matcher{}
+
+func Match(expected interface{}) *Matcher {
```
khuddlefish: It's a bit confusing to have both this and the other `Match` function called the same thing. Can we think of another name? MatchStruct?

Or maybe we could make separate matchers for `MatchIncluding` or `MatchExcluding`?

I'm looking at [gstruct](https://github.com/onsi/gomega/tree/master/gstruct) for examples

jmcatamney: Yeah, the namespace collision there is a bit annoying.  `MatchStruct` is probably the simplest change, but `structmatcher.MatchStruct` has a bit of a stutter.  Splitting it into `MatchIncluding`, `MatchExcluding`, and `MatchExactly` would also work.  No strong opinions on which approach to take.

structmatcher/structmatcher.go
```diff
@@ -87,3 +87,48 @@ func ExpectStructsToMatchIncluding(expected interface{}, actual interface{}, inc
...
+	} else if m.excludingFields != nil {
+		m.mismatches = StructMatcher(m.expected, actual, true, false, m.excludingFields...)
+	} else {
+		m.mismatches = StructMatcher(m.expected, actual, false, false)
+	}
+	return len(m.mismatches) == 0, nil
+}
+
+func (m *Matcher) FailureMessage(actual interface{}) (message string) {
+	//return format.Message(actual, "to match struct", m.expected)
```
jmcatamney, line 121: Looks like the comment is outdated and should be deleted.

jmcatamney:
Other than the `Match` name issue, functionally everything else looks good to me.

PR #48, opened 2019-11-13
recursive structmatcher

dsharp-pivotal:
This could potentially break some existing tests if they look at individual messages, but that seems an unlikely way to use structmatcher.


jmcatamney (approved):
LGTM.

Thanks for doing this, we'd been punting on implementing recursive matching for quite a while.

dsharp-pivotal:
This doesn't work when some of the deeply nested structs have private fields. Calling .Interface() on a struct with private fields is not allowed apparently.

dsharp-pivotal:
Okay. I think this is good now. PTAL at the last commit.

jmcatamney (approved):

PR #49, opened 2019-12-02
Fix gofmt detectecd error found during `make test`.

edespino:
The `gp-common-go-libs` pipeline executing `! gofmt -l structmatcher/ | read` during the `lint` make target has identified this as a failure. This update corrects the failure.

This was introduced with https://github.com/greenplum-db/gp-common-go-libs/commit/7f408d257e14cc8eefdfc111deea7cd930e29c17


khuddlefish (approved):
Oops

PR #51, opened 2019-12-19
Migrate to go mod and golangci-lint

magi345:
Deprecated
 - dep
 - gpmetalinter

Migrated to
 - go mod
 - golangci-lint


magi345:
CI see: https://prod.ci.gpdb.pivotal.io/teams/main/pipelines/gp-common-go-libs-wh/jobs/units/builds/14

yangru (approved):
LGTM

magi345:
@chrishajas @khuddlefish @pf-qiu @fanfuxiaoran @dsharp-pivotal 
I have this PR to do housekeeping for gp-common-go-libs, please review if these changes ok to your projects.
I will hold merge until 2020-01-07. Please comment if any concerns or impacts.

- Hao

pf-qiu:
We don't use master branch so shouldn't be a problem for us.

magi345:
> We don't use master branch so shouldn't be a problem for us.

Maybe separate issue, just curious why needs to maintain your code in another branch?
Is there any conflict with anybody if merge your contribution to master?

Makefile
```diff
@@ -14,10 +14,10 @@ GOFLAGS :=
 dependencies :
 		go get github.com/alecthomas/gometalinter
 		gometalinter --install
-		curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
-		dep ensure
-		@cd vendor/golang.org/x/tools/cmd/goimports; go install .
-		@cd vendor/github.com/onsi/ginkgo/ginkgo; go install .
+		go get golang.org/x/tools/cmd/goimports
+		go get github.com/onsi/ginkgo/ginkgo
+		go mod vendor
```
dsharp-pivotal: There shouldn't be any need for a vendor directory with go modules. Did you have a specific need for it?

Makefile
```diff
@@ -14,10 +14,10 @@ GOFLAGS :=
...
+		go mod tidy
```
dsharp-pivotal, line 18: This also seems unnecessary. It could cause unexpected diffs.

Makefile
```diff
@@ -14,10 +14,10 @@ GOFLAGS :=
 dependencies :
 		go get github.com/alecthomas/gometalinter
 		gometalinter --install
-		curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
-		dep ensure
-		@cd vendor/golang.org/x/tools/cmd/goimports; go install .
-		@cd vendor/github.com/onsi/ginkgo/ginkgo; go install .
+		go get golang.org/x/tools/cmd/goimports
+		go get github.com/onsi/ginkgo/ginkgo
```
dsharp-pivotal, line 17: Should these be `go install`?

Makefile
```diff
@@ -12,8 +12,7 @@ GOFLAGS :=
 .PHONY : coverage
 
 dependencies :
-		go get github.com/alecthomas/gometalinter
-		gometalinter --install
+		curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.21.0
```
dsharp-pivotal: FWIW, we install with `go install -i github.com/golangci/golangci-lint/cmd/golangci-lint`

Makefile
yangru: Hi, @dsharp-pivotal
From the best practise, It's always recommended to run `go mod tidy` after go install or go build. 

Here's some explanation from golang repo:
https://github.com/golang/go/issues/27633

_The reason is that go mod tidy is agnostic of GOOS/GOARCH or build tags; go build is lazy and only does the minimal amount of work required for the current GOOS/GOARCH and build tags._

_cc @bcmills - I wonder whether we should update the documentation for go mod tidy to include some note about this operation being agnostic of GOOS/GOARCH and build tags:_
...
_Because one reading of the documentation is that go build and go mod tidy should leave go.mod in the same state._

Makefile
magi345: `go get github.com/onsi/ginkgo/ginkgo` will install the binary executable into $GOPATH/bin/ so don't need to run `go install`.
The same with goimports.

Makefile
magi345: According to the readme of golangci-lint https://github.com/golangci/golangci-lint#binary, it suggests use curl to install.
It says `Please, do not install golangci-lint by go get:`

```
1. go.mod replacement directive doesn't apply. It means you will be using patched version of golangci-lint.
2. it's much slower than binary installation
3. its stability depends on your Go version (e.g. on this compiler Go <= 1.12 bug).
4. it's not guaranteed to work: e.g. we've encountered a lot of issues with Go modules hashes.
5. it allows installation from master branch which can't be considered stable.
```

I think `go install` is the same with with `go get`, it will pollute the go.mod/go.sum files.

Makefile
magi345: I removed the `go mod vendor` statement

Makefile
magi345: I kept the `go mod tidy` statement according to above explanation from yangru

dsharp-pivotal (approved):

PR #54, opened 2020-03-05
dd cmd

yangru:
This is the tool we are currently using in our repo to get bytes at specified positions in the gpdb log file iteratively.
At first, we wanted to use the original dd command, but we found centos6 can't satisfy our needs, and to make it compatible with all OSes, we built this dd tool ourselves.

Usage:
gpdd -if [input FILE] -of [output FILE] -skip [skip bytes] -count [bytes to copy]

Examples:
Copy 100 bytes from a.in to b.out, skipping the starting 20 bytes:
	gpdd -if a.in -of b.out -skip 20 -count 100

Print the beginning 100 bytes from a.in to stdout:
	gpdd -if a.in -count 100

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ lint :
 		golangci-lint run -v
 
 unit :
-		ginkgo -r -keepGoing -randomizeSuites -randomizeAllSpecs cluster dbconn gplog iohelper structmatcher conv 2>&1
+		ginkgo -r -keepGoing -randomizeSuites -randomizeAllSpecs cluster dbconn gplog iohelper structmatcher conv cmd/... 2>&1
 
 test : lint unit
 
--- a/cmd/dd/dd.go
+++ b/cmd/dd/dd.go
@@ -22,12 +22,12 @@ Print the beginning 100 bytes from a.in to stdout:
 var (
 	// main operation modes
 	ifPath = flag.String("if", "", "read from FILE")
-	skip = flag.Int64("skip", 0, "skip N bytes at start of input FILE, default is 0")
-	count = flag.Int64("count", 0, "copy only N bytes, default is 0")
+	skip   = flag.Int64("skip", 0, "skip N bytes at start of input FILE, default is 0")
+	count  = flag.Int64("count", 0, "copy only N bytes, default is 0")
 	ofPath = flag.String("of", "", "write to FILE, default is stdout")
 )
 
-var exitCode   = 0
+var exitCode = 0
 
 func usage() {
 	fmt.Fprintln(os.Stderr, usageMessage)
@@ -78,9 +78,6 @@ func report(err error) {
 }
 
 func main() {
-	// call ddMain in a separate function
-	// so that it can use defer and have them
-	// run before the exit.
 	ddMain()
 	os.Exit(exitCode)
 }
--- a/cmd/dd/dd.go
+++ b/cmd/dd/dd.go
@@ -1,0 +1,128 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io"
+	"os"
+)
+
+const usageMessage = "" +
+	`Usage:
+dd -if [input FILE] -of [output FILE] -skip [skip bytes] -count [bytes to copy]
+
+Examples:
+Copy 100 bytes from a.in to b.out, skipping the starting 20 bytes:
+	dd -if a.in -of b.out -skip 20 -count 100
+
+Print the beginning 100 bytes from a.in to stdout:
+	dd -if a.in -count 100
+`
+
+var (
+	// main operation modes
+	ifPath = flag.String("if", "", "read from FILE")
+	skip = flag.Int64("skip", 0, "skip N bytes at start of input FILE, default is 0")
+	count = flag.Int64("count", 0, "copy only N bytes, default is 0")
+	ofPath = flag.String("of", "", "write to FILE, default is stdout")
+)
+
+var exitCode   = 0
+
+func usage() {
+	fmt.Fprintln(os.Stderr, usageMessage)
+	fmt.Fprintln(os.Stderr, "Flags:")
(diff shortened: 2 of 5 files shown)


pivotal-issuemaster:
@yangru Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/greenplum?repositoryId=greenplum-db/gp-common-go-libs&pullRequestId=54)!

[Click here](https://cla.pivotal.io/sync/greenplum?repositoryId=greenplum-db/gp-common-go-libs&pullRequestId=54) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@yangru Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/greenplum?repositoryId=greenplum-db/gp-common-go-libs&pullRequestId=54)!

jmcatamney:
Could you explain your reasoning for wanting to add this to `gp-common-go-libs` and what problem you were trying to solve?  I don't think this PR is a good addition to this repo, for a few reasons:

1) `gp-common-go-libs` is intended to be a set of libraries to be imported while this code compiles a separate `dd` binary, so it doesn't fit from a style and usage perspective.

2) The copyright statement in the test file and the divergent coding style implies that this code is copy-pasted from elsewhere, so there's no reason to add that to this repo when one can use the original code as-is.

3) `dd` comes standard with most Unix-like OSes, while the `dd.exe` mention in the test file implies this is meant to be a reimplementation for the benefit of Windows code.  As Greenplum utilities only support environments where `dd` is already present, there's no need for this addition.

yangru:
Hi, @jmcatamney , 
Thank you for your review!
This is my reply:

> Could you explain your reasoning for wanting to add this to `gp-common-go-libs` and what problem you were trying to solve? I don't think this PR is a good addition to this repo, for a few reasons:

This is the tool we are currently using in our repo to get bytes at specified positions in the gpdb log file iteratively.
 At first, we wanted to use the original dd command, but we found centos6 can't satisfy our needs,  and to make it compatible with all OSes, we built this dd tool ourselves. 
We don't know whether there're other gpdb users struggling with this problem, so we decided to open source this tool.

> 
> 1. `gp-common-go-libs` is intended to be a set of libraries to be imported while this code compiles a separate `dd` binary, so it doesn't fit from a style and usage perspective.

I see your point, but users can import it, build it and run it. That's simple to use and proved robust. Maybe this is a good start to include some useful tools. 

> 2. The copyright statement in the test file and the divergent coding style implies that this code is copy-pasted from elsewhere, so there's no reason to add that to this repo when one can use the original code as-is.

I borrowed some utility functions from the core go library to get go command path,  I see that's not proper, so I replaced them with a simpler implementation.

> 3. `dd` comes standard with most Unix-like OSes, while the `dd.exe` mention in the test file implies this is meant to be a reimplementation for the benefit of Windows code.  As Greenplum utilities only support environments where `dd` is already present, there's no need for this addition.

This command can also run in windows. So I added this .exe suffix.
But to make things clear, I've changed code a bit to separate windows and others.
Unix's spirit is to try our best to make things compatible, and that's a test file, so I think it is not a big problem.

jmcatamney:
Thanks for the additional context, @yangru, that definitely helps.

> This is the tool we are currently using in our repo to get bytes at specified positions in the gpdb log file iteratively.

It seems to me that if this is what you're ultimately trying to do, a package to specifically do _that_ would be a better fit for this repo than a generic `dd` reimplementation.  Parsing GPDB log files is something a lot of users might want to do and a library to handle that for them would be a better value-add, in the same way that e.g. `cluster` is specifically aimed to make it easier to work with a GPDB cluster instead of being just a generic ssh-on-arbitrary-hosts package.  And integrating things like that would probably have less overhead than shelling out to an executable, as well.

Is your current logic to deal with the log files sufficiently generic and modular that you could split it into its own package and put that into a PR?

yangru:
Hi, @jmcatamney ,
Thank you for your interest!
The work we did for harvesting pglog is more like an integrated **service**, it needs to start a daemon, create several db tables(one of them is to store log data) and maintain them(vacuum, analyze, create partitions and etc...), create some customized tools(like dd). It can't fit into or the **library** scope very well.
And our team maintains a close source project, we have to be careful about and get approval before the exposure of the business logic. 
However, dd command is simple and is an out-of-the-box tool for gp users.
I'm grateful that you read my explanation carefully and help to find something really valuable. Our team will extract more tools and libraries to this repo later! 

> Thanks for the additional context, @yangru, that definitely helps.
> 
> > This is the tool we are currently using in our repo to get bytes at specified positions in the gpdb log file iteratively.
> 
> It seems to me that if this is what you're ultimately trying to do, a package to specifically do _that_ would be a better fit for this repo than a generic `dd` reimplementation. Parsing GPDB log files is something a lot of users might want to do and a library to handle that for them would be a better value-add, in the same way that e.g. `cluster` is specifically aimed to make it easier to work with a GPDB cluster instead of being just a generic ssh-on-arbitrary-hosts package. And integrating things like that would probably have less overhead than shelling out to an executable, as well.
> 
> Is your current logic to deal with the log files sufficiently generic and modular that you could split it into its own package and put that into a PR?

chrishajas:
My main concern here is that it builds a binary. Currently, this repo is a developer library. Nobody is responsible for packaging/shipping anything in it to a user or via a release. It's more providing common functions for go code for development and is consumed automatically from other applications via `go gets`. If you have to call this binary for your application, I think it would be best if it exists either as a binary in the gpdb repo (or your repo), or as a library in this repo where your repo provides a wrapper and handles packaging/shipping.

yangru:
Hi, @chrishajas and @jmcatamney , 
I see your concerns and worries,
will follow this repo's rules and close this pull request.
I'll contribute to this repo once there're other good candidates in the future.

PR #55, opened 2020-03-06
cluster: Prevent panic when executing commands

oliverralbertini:
The cluster.Executor should be set when newing up a cluster object.

Authored-by: Oliver Albertini <oalbertini@pivotal.io>


jmcatamney (approved):
Whoops, looks like I accidentally deleted that line when I refactored the function.  Good catch!

PR #59, opened 2020-10-05
Update pgx to v4.9.0 and disable statement caching

kyeap-vmware:
pgx v4.9.0 exposes a channel that can be used to tell when a query
canceled via golang's context is finished in postgres. This is useful
for cleanup when a program has been interrupted and need to cancel
blocked queries.  Disabled pgx's automatic statement caching because it
causes invalid caching in GPDB4 when creating, destroying, and creating
the same object again on the same connection.

Authored-by: Kevin Yeap <kyeap@vmware.com>

### DIFF
--- a/dbconn/dbconn.go
+++ b/dbconn/dbconn.go
@@ -194,8 +194,14 @@ func (dbconn *DBConn) Connect(numConns int) error {
 	if dbconn.ConnPool != nil {
 		return errors.Errorf("The database connection must be closed before reusing the connection")
 	}
-	// This string takes in the literal user/database names. They do not need to be escaped or quoted.
-	connStr := fmt.Sprintf("postgres://%s@%s:%d/%s?sslmode=disable", dbconn.User, dbconn.Host, dbconn.Port, dbconn.DBName)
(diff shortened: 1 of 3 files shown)


shivzone (approved):

dbconn/dbconn.go
```diff
@@ -194,8 +194,15 @@ func (dbconn *DBConn) Connect(numConns int) error {
 	if dbconn.ConnPool != nil {
 		return errors.Errorf("The database connection must be closed before reusing the connection")
 	}
-	// This string takes in the literal user/database names. They do not need to be escaped or quoted.
-	connStr := fmt.Sprintf("postgres://%s@%s:%d/%s?sslmode=disable", dbconn.User, dbconn.Host, dbconn.Port, dbconn.DBName)
+	// This string takes in the literal user/database names. They do not need
+	// to be escaped or quoted.
+	// By default pgx/v4 by default turns on automatic prepared statement
```
jimmyyih: "By default" twice here.  Remove the second one... unless this is a super default then it's okay. 😄

jimmyyih (approved):

PR #63, opened 2022-01-26
structmatcher: Use MatchStruct matcher in deprecated implementations

dsharp-pivotal:
We're trying to upgrade to ginkgo/v2, but structmatcher imports ginkgo v1. We can avoid the issue by not importing ginkgo at all. In this case, structmatcher only called Fail() in the deprecated ExpectXxx() functions. These can be replaced with using the MatchStruct() GomegaMatcher.

More info: https://github.com/onsi/ginkgo/issues/875

### DIFF
--- a/structmatcher/structmatcher.go
+++ b/structmatcher/structmatcher.go
@@ -9,7 +9,6 @@ import (
 	"reflect"
 	"strings"
 
-	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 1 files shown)


jimmyyih (approved):

dsharp-pivotal:
@jimmyyih If this looks good, can you merge it? I don't have write access to this repository (anymore?).

PR #64, opened 2022-04-20
Remove ginkgo v1 reference from testhelper package

jimmyyih:
The ginkgo v1 reference in testhelper's function.go was making the
testhelper package incompatible with ginkgo v2. Remove the ginkgo v1
reference and replace the Fail() calls with equivalent gomega Expect
calls.


jimmyyih:
This goes along the same lines as https://github.com/greenplum-db/gp-common-go-libs/pull/63.  With that PR and this PR, gp-common-go-libs test packages should be compatible with ginkgo v2.

shivzone (approved):

PR #65, opened 2022-04-27
dbconn: add the ability to connect in utility mode

jmcatamney:
This commit modifies the dbconn.Connect function to take a boolean
which, if set to true, connects in utility mode.  It also adds some
helper functions for more explicit invocation.

In order to test this functionality, it was necessary to modify the
TestDriver struct to be able to return different errors on different
calls, so this commit also makes that change.

### DIFF
--- a/dbconn/dbconn.go
+++ b/dbconn/dbconn.go
@@ -64,7 +64,7 @@ type DBDriver interface {
 type GPDBDriver struct {
 }
 
-func (driver GPDBDriver) Connect(driverName string, dataSourceName string) (*sqlx.DB, error) {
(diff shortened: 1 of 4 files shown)


jimmyyih (approved):
From just looking at the patch, it seems like you were very careful to not affect current usages of the dbconn Connect function API.  Is that correct?  Otherwise, the patch looks good.  That scenario about which utility mode GUC to use is definitely annoying but I think you handled it pretty well.  Looks fine to me. 👍

jmcatamney:
> From just looking at the patch, it seems like you were very careful to not affect current usages of the dbconn Connect function API. Is that correct?

Yep, the existing functionality is unchanged for non-utility connections.

PR #67, opened 2022-07-15
add ssl support

xiaoxiaoHe-E:
Ssl connection is supported by this version of lib/pq it used. But in gp-common-go-libs, when connect to gpdb, sslmode is set to disable in hard code. 

We want it to read sslmode from the environment variable. Just remove the hard code `sslmode="disable"` will work.

### DIFF
--- a/dbconn/dbconn.go
+++ b/dbconn/dbconn.go
@@ -193,7 +193,8 @@ func (dbconn *DBConn) Connect(numConns int) error {
 	if krbsrvname == "" {
 		krbsrvname = "postgres"
 	}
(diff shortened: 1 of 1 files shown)


hyongtao-code (approved):
Looks good.

beeender (commented):
LGTM

Are changes in `go.mod` and `go.sum` necessary? Will it break others who is using the target branch?

xiaoxiaoHe-E:
I added them just to replace the repo in my local workspace. I'm not sure if it will break others' work. I'll remove it if it may introduce problems.

huiliang-liu (approved):

beeender (approved):

PR #68, opened 2022-08-11
Add "Completed" field to ShellCommand

kaknikhil:
Added this explicit field to make it easier to determine if a command has completed its execution regardless of success or failure.


jimmyyih (approved):
Currently, there is no easy way to tell if a command in a commandList has completed or not.  If the command completed with an error, we can easily check for `Error != nil` but a command completed successfully cannot be determined by the Stdout since it defaults to an empty string prior to executing the command which is also a valid output of a completed command. Because of that, having a Completed variable in the struct would be useful for applications that need to do asynchronous checking of the command list. Looks good to me; and it doesn't look like it would break any existing applications that use the cluster package.

If there are no objections to this change within a couple days, I'll merge this PR and create a new tag.

PR #72, opened 2022-12-20
Change uses of "master" to "coordinator"

jmcatamney:
As part of an effort to use more inclusive terminology, GPDB 7 is renaming the GPDB "master" to the "coordinator" and the "master" branch to "main", and any related code is to be changed as well.

This commit adds new INCLUDE_COORDINATOR and EXCLUDE_COORDINATOR scopes as synonyms of the existing INCLUDE_MASTER and EXCLUDE_MASTER scopes, instead of renaming them, to maintain backwards compatibility for now. All internal usage of the term is updated, however.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,5 +61,5 @@ set-prod:
 	fly --target prod set-pipeline --check-creds \
 	--pipeline=gp-common-go-libs \
 	-c ci/pipeline.yml \
-	--var=branch=master \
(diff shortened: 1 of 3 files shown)


AJR-VMware (approved):
Looks good to me.

I take it there wasn't much need for conditionally printing out one term vs another in the common libs?

PR #77, opened 2023-04-13
Miscellaneous mirror-related enhancements

jmcatamney:
This commit changes cluster behavior in the following ways in order to make it easier to work with mirrors:

1) The ContentIDs array is now deduplicated and sorted in the Cluster constructor, so that a cluster with mirrors won't have duplicate values in that array and segments can be added in any order without affecting the order of ContentIDs.

2) GetSegmentConfiguration can now retrieve only mirror information if desired, so that if some code wants to work with primaries and mirrors separately it can do so without having to retrieve both primary and mirror information at once and then manually separate it out into two Clusters.

3) TestExecutor now supports returning different values on subsequent calls for both ExecuteLocalCommand and ExecuteClusterCommand, in order to make it easier to test functions that might make one call to deal with primaries and another to deal with mirrors (or any other use case involving multiple calls per function, of course).


AJR-VMware (approved):
No notes, LGTM!

PR #80, opened 2023-08-30
dbconn: Add SelectInt and SelectIntSlice

jmcatamney:
Selecting ints can have the same issues as selecting strings (may be NULL, may not exist, needs a destination struct, etc.) so for user convenience this commit adds SelectInt and SelectIntSlice functions that work essentially like SelectString and SelectStringSlice.


AJR-VMware (approved):
LGTM!

PR #83, opened 2023-10-20
Move error formatting from GPVO codebase

djianwen-vmware:
Created PR again to omit internal story number.

### DIFF
--- a/gperror/gperror_test.go
+++ b/gperror/gperror_test.go
@@ -1,0 +1,55 @@
+package gperror_test
+
+import (
+	"errors"
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/greenplum-db/gp-common-go-libs/gperror"
+)
+
+func TestGpError(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Gperror Suite")
+}
+
+var _ = Describe("gperror", func() {
+	var testErr *gperror.GpError
+
+	BeforeEach(func() {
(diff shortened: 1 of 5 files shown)


stuart-pollock:
You should add "gperror" to `show_coverage.sh`.

go.mod
```diff
@@ -7,37 +7,29 @@ require (
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/jackc/pgx/v4 v4.18.1
 	github.com/jmoiron/sqlx v1.3.5
-	github.com/onsi/ginkgo v1.16.5
+	github.com/onsi/ginkgo/v2 v2.8.4
```
stuart-pollock: Why was the version of Ginkgo bumped from v1 to v2?

stuart-pollock (dismissed):

gperror/gperror_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	BeforeEach(func() {
+		testErr = &gperror.GpError{
+			ErrorCode: gperror.ErrorCode(4321),
+			Err:       errors.New("test-error"),
+		}
+	})
+
+	Context("Error", func() {
+		When("the function is called", func() {
```
jmcatamney: Stylistically, the tests for the other packages use `Describe` instead of `Context`, and don't use `When` because it doesn't generally add any useful information for simple unit tests.  If you can tweak these tests to match, that would be great.

Everything looks good otherwise.

tatb-vmware: Addressed in most recent commit.

jmcatamney (approved):

PR #85, opened 2024-01-05
Introducing API GetSegmentConfigurationFromFile() for configuration retrieval from file

shrakesh:
This PR introduces an API GetSegmentConfigurationFromFile() to parse the gpsegconfig_dump file to retrieve segment configuration information.
The recommended use of the API is to get the contents of gp_segment_configuration when the database is down.

In the realm of gpdb, the gpsegconfig_dump file holds the crucial segment configuration information. this file is generated in the coordinator
data directory through the fts process. gpsegconfig_dump is always in sync with gp_segment_configuration as on each fts probe the file
gets updated if there is any change in the segment configuration. Various fts gucs govern the frequency of writing to this file.

Note: Since the gpsegconfig_dump file is updated by fts process the information returned by
this function can be a bit stale since the user can configure fts to run less frequently

The gpsegconfig_dump file follows a structured format, as illustrated in the example below:
```
1 -1 p p n u 6000 localhost localhost /data/temp1
2 0 p p n u 6002 localhost localhost /data/temp2
3 1 p p n u 6003 localhost localhost /data/temp3
4 2 p p n u 6004 localhost localhost /data/temp4
```

Example Usage:
```
   segments, err := GetSegmentConfigurationFromFile("/path/to/coordinator/data/dir")
   if err != nil {
       //Handle error
       return
   }
   
```
*. if gpsegconfig_dump has the following content ( with data-dir).
```
   1 -1 p p n u 6000 localhost localhost /data/qddir
   2 0 p p n u 6002 localhost localhost /data/seg1
```
   SegConfig will have the DataDir field populated
*. gpsegconfig_dump has the following content ( without data-dir)
```
    1 -1 p p n u 6000 localhost localhost
    2 0 p p n u 6002 localhost localhost
```
    SegConfig will have the DataDir field empty

The structured data, captured in the gpsegconfig_dump file, is now accessible through the newly added API GetSegmentConfigurationFromFile().
This enhancement helps the other utilities to perform some operations even if the database is offline.

e.g.
In gpsupport utility if the database is down we can use this API to retrieve the data dirs to perform a log collection.
or
In gpdeletesystem if the database is down the API can be used to perform the database cleanup still.

The reason to add it in gp-common-go-libs is that this is the growing API library that is being used by almost all the newly created gp modern utilities like project spine/ gpdr/ gpupgrade/ gpbackup/gprestore.

Added test cases to do the unit testing around the following scenario
when the file is valid it should provide valid results.
when the file is invalid it should fail in parsing or reading it.
added test cases to cover old and new field counts of gpsegconfig_dump. 
when the user passes the wrong coordinatorDataDir argument.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -46,12 +46,16 @@ type Cluster struct {
 }
 
 type SegConfig struct {
-	DbID      int
-	ContentID int
-	Role      string
-	Port      int
-	Hostname  string
-	DataDir   string
+	DbID          int
+	ContentID     int
+	Role          string
+	PreferredRole string
+	Mode          string
+	Status        string
+	Port          int
+	Hostname      string
+	Address       string
+	DataDir       string
 }
 
 /*
@@ -522,8 +526,12 @@ SELECT
 	s.dbid,
 	s.content as contentid,
 	s.role,
+	s.preferred_role as preferredrole,
+	s.mode,
+	s.status,
 	s.port,
 	s.hostname,
+	s.address,
 	e.fselocation as datadir
 FROM gp_segment_configuration s
 JOIN pg_filespace_entry e ON s.dbid = e.fsedbid
@@ -542,8 +550,12 @@ SELECT
 	dbid,
 	content as contentid,
 	role,
+	preferred_role as preferredrole,
+	mode,
+	status,
 	port,
 	hostname,
+	address,
 	datadir
 FROM gp_segment_configuration
 %s
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -6,10 +6,14 @@ package cluster
  */
 
 import (
+	"bufio"
 	"bytes"
 	"fmt"
+	"os"
 	"os/exec"
+	"path"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/greenplum-db/gp-common-go-libs/dbconn"
@@ -574,4 +578,126 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn, getMirrors ...bool) 
 	segConfigs, err := GetSegmentConfiguration(connection, len(getMirrors) == 1 && getMirrors[0])
 	gplog.FatalOnError(err)
 	return segConfigs
+}
+
+/*GetSegmentConfigurationFromFile parse the gpsegconfig_dump file to retrieve segment configuration information.
+Recommended use of the api is to get the contents of gp_segment_configuration when database is down.
+If the database is up, use GetSegmentConfiguration()/MustGetSegmentConfiguration() instead.
+
+gpsegconfig_dump file gets created at $COORDINATOR_DATA_DIR/gpsegconfig_dump by fts process
+The frequency of writing to this file is governed by various fts gucs.
+
+Note: Since the gpsegconfig_dump file is updated by fts process the information returned by
+this function can be a bit stale since user can configure fts to run less frequently
+
+The gpsegconfig_dump file follows a structured format, as illustrated in the example below:
+1 -1 p p n u 6000 localhost localhost /data/temp1
+2 0 p p n u 6002 localhost localhost /data/temp2
+3 1 p p n u 6003 localhost localhost /data/temp3
+4 2 p p n u 6004 localhost localhost /data/temp4
+
+Example Usage:
+   segments, err := GetSegmentConfigurationFromFile("/path/to/coordinator/data/dir")
+   if err != nil {
+       //Handle error
+       return
+   }
+
+*. if gpsegconfig_dump have the following content ( with data-dir).
+   1 -1 p p n u 6000 localhost localhost /data/qddir
+   2 0 p p n u 6002 localhost localhost /data/seg1
+   SegConfig will have DataDir field populated
+
+*. gpsegconfig_dump has following content ( without data-dir)
+	1 -1 p p n u 6000 localhost localhost
+    2 0 p p n u 6002 localhost localhost
+    SegConfig will have DataDir field empty
+
+
+Parameters:
+  -  coordinatorDataDir - The path to the coordinator data directory containing gpsegconfig_dump file.
+     can be retrieved from env var COORDINATOR_DATA_DIRECTORY
+     (e.g. /Users/shrakesh/workspace/gpdb/gpAux/gpdemo/datadirs/qddir/demoDataDir-1)
+
+Returns:
+  - []SegConfig: A slice of SegConfig structures representing the segment configuration.
+  - error: If any occurs during file reading and parsing.
+*/
+
+func GetSegmentConfigurationFromFile(coordinatorDataDir string) ([]SegConfig, error) {
+
+	/*Check if the given argument coordinator_data_dir is empty*/
+	if len(strings.TrimSpace(coordinatorDataDir)) == 0 {
+		return nil, fmt.Errorf("Coordinator data directory path is empty")
+	}
+
+	/*Generate gpsegconfig_dump file path*/
+	gpsegconfigDump := path.Join(coordinatorDataDir, "gpsegconfig_dump")
+
+	/* Open gpsegconfig_dump */
+	fd, err := os.Open(gpsegconfigDump)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to open file %s. Error: %s", gpsegconfigDump, err.Error())
+	}
+	defer fd.Close()
+
+	results := make([]SegConfig, 0)
+	scanner := bufio.NewScanner(fd)
+
+	/*scanning file line by line to extract the fields into SegConfig struct*/
+	for scanner.Scan() {
+		fields := strings.Fields(scanner.Text())
+		parts := len(fields)
+
+		/* older version of gpsegconfig_dump has 9 parts as it doesn't have datadir
+			1 -1 p p n u 7000 shrakeshSMD6M.vmware.com shrakeshSMD6M.vmware.com
+		newer version of gpsegconfig_dump has 10 parts as it does have datadir
+			1 -1 p p n u 7000 shrakeshSMD6M.vmware.com shrakeshSMD6M.vmware.com /data/qddir/demoDataDir-1 */
+		if parts != 9 && parts != 10 {
+			return nil, fmt.Errorf("Unexpected number of fields (%d) in line: %s", parts, scanner.Text())
+		}
+
+		dbID, err := strconv.Atoi(fields[0])
+		if err != nil {
+			return nil, fmt.Errorf("Failed to convert dbID with value %s to an int. Error: %s", fields[0], err.Error())
+		}
+
+		content, err := strconv.Atoi(fields[1])
+		if err != nil {
+			return nil, fmt.Errorf("Failed to convert content with value %s to an int. Error: %s", fields[1], err.Error())
+		}
+
+		port, err := strconv.Atoi(fields[6])
+		if err != nil {
+			return nil, fmt.Errorf("Failed to convert port with value %s to an int. Error: %s", fields[6], err.Error())
+		}
+
+		// there are 10 fields in new version of gpsegconfig_dump file
+		datadir := ""
+		if parts == 10 {
+			datadir = fields[9]
+		}
+
+		seg := SegConfig{
+			DbID:          dbID,
+			ContentID:     content,
+			Role:          fields[2],
+			PreferredRole: fields[3],
+			Mode:          fields[4],
+			Status:        fields[5],
+			Port:          port,
+			Hostname:      fields[7],
+			Address:       fields[8],
+			DataDir:       datadir,
+		}
+
+		results = append(results, seg)
+	}
+
+	/* validating error during gpsegconfig_dump file read */
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("Failed to read gpsegconfig_dump file %s: %s", gpsegconfigDump, err.Error())
+	}
+
+	return results, nil
 }
--- a/cluster/cluster_test.go
+++ b/cluster/cluster_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/user"
+	"path"
 	"testing"
 
 	sqlmock "github.com/DATA-DOG/go-sqlmock"
@@ -35,6 +36,17 @@ func expectPathToExist(path string) {
 	if _, err := os.Stat(path); os.IsNotExist(err) {
 		Fail(fmt.Sprintf("Expected %s to exist", path))
 	}
+}
+
+func createSegConfigFile(content string) *os.File {
+	filename := path.Join(os.TempDir(), "gpsegconfig_dump")
+	confFile, err := os.OpenFile(filename, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0600)
+	Expect(err).To(BeNil())
+	_, err = confFile.WriteString(content)
+	Expect(err).To(BeNil())
+
+	defer confFile.Close()
+	return confFile
 }
 
 var _ = BeforeSuite(func() {
@@ -75,6 +87,180 @@ var _ = Describe("cluster/cluster tests", func() {
 			Expect(cmd).To(Equal([]string{"ssh", "-o", "StrictHostKeyChecking=no", "testUser@some-host", "ls"}))
 		})
 	})
+
+	Describe("GetSegmentConfigurationFromFile", func() {
+		It("should return expected result for a new (10 fields) gpsegconfig_dump file", func() {
+			//create temp file with the sample data from new version
+			expRes := cluster.SegConfig{
+				DbID:          1,
+				ContentID:     -1,
+				Role:          "p",
+				PreferredRole: "p",
+				Mode:          "n",
+				Status:        "u",
+				Port:          7000,
+				Hostname:      "localhost",
+				Address:       "localhost",
+				DataDir:       "/data/qddir/demoDataDir-1",
+			}
+			content := fmt.Sprintf("%d %d %s %s %s %s %d %s %s %s", expRes.DbID, expRes.ContentID, expRes.Role, expRes.PreferredRole, expRes.Mode, expRes.Status, expRes.Port, expRes.Hostname, expRes.Address, expRes.DataDir)
+			tempConfFile := createSegConfigFile(content)
+
+			//call the function under test
+			result, err := cluster.GetSegmentConfigurationFromFile(os.TempDir())
+			Expect(err).To(BeNil())
+			Expect(result).To(HaveLen(1))
+			Expect(result[0]).To(Equal(expRes))
+
+			//Cleanup
+			os.Remove(tempConfFile.Name())
+		})
+
+		It("should return expected result for an old (9 fields) gpsegconfig_dump file", func() {
+			//create temp file with the sample data from new version
+			expRes := cluster.SegConfig{
+				DbID:          1,
+				ContentID:     -1,
+				Role:          "p",
+				PreferredRole: "p",
+				Mode:          "n",
+				Status:        "u",
+				Port:          7000,
+				Hostname:      "localhost",
+				Address:       "localhost",
+			}
+			content := fmt.Sprintf("%d %d %s %s %s %s %d %s %s", expRes.DbID, expRes.ContentID, expRes.Role, expRes.PreferredRole, expRes.Mode, expRes.Status, expRes.Port, expRes.Hostname, expRes.Address)
+			tempConfFile := createSegConfigFile(content)
+
+			//call the function under test
+			result, err := cluster.GetSegmentConfigurationFromFile(os.TempDir())
+			Expect(err).To(BeNil())
+			Expect(result).To(HaveLen(1))
+			Expect(result[0]).To(Equal(expRes))
+
+			//Cleanup
+			os.Remove(tempConfFile.Name())
+		})
+
+		It("should return expected result for multiline gpsegconfig_dump file", func() {
+			//create temp file with the sample data from new version
+			expRes := []cluster.SegConfig{
+				{
+					DbID:          1,
+					ContentID:     -1,
+					Role:          "p",
+					PreferredRole: "p",
+					Mode:          "n",
+					Status:        "u",
+					Port:          7000,
+					Hostname:      "localhost",
+					Address:       "localhost",
+					DataDir:       "/data/qddir/demoDataDir-1",
+				},
+				{
+					DbID:          2,
+					ContentID:     -1,
+					Role:          "m",
+					PreferredRole: "m",
+					Mode:          "n",
+					Status:        "u",
+					Port:          7001,
+					Hostname:      "localhost",
+					Address:       "localhost",
+					DataDir:       "/data/standby/demoDataDir-2",
+				},
+			}
+			var content string
+			for _, segconf := range expRes {
+				text := fmt.Sprintf("%d %d %s %s %s %s %d %s %s %s\n", segconf.DbID, segconf.ContentID, segconf.Role, segconf.PreferredRole, segconf.Mode, segconf.Status, segconf.Port, segconf.Hostname, segconf.Address, segconf.DataDir)
+				content = content + text
+			}
+
+			tempConfFile := createSegConfigFile(content)
+
+			//call the function under test
+			result, err := cluster.GetSegmentConfigurationFromFile(os.TempDir())
+			Expect(err).To(BeNil())
+			Expect(result).To(HaveLen(2))
+			Expect(result).To(Equal(expRes))
+
+			//Cleanup
+			os.Remove(tempConfFile.Name())
+		})
+
+		It("should fail when empty coordinator data directory is provided to function", func() {
+			// Call the function under test
+			result, err := cluster.GetSegmentConfigurationFromFile("")
+
+			// Assertions
+			Expect(err).To(HaveOccurred())
+			Expect(result).To(BeNil())
+			Expect(err.Error()).To(ContainSubstring("Coordinator data directory path is empty"))
+		})
+
+		It("should fail when reading invalid file/path", func() {
+			// Call the function under test
+			result, err := cluster.GetSegmentConfigurationFromFile("/tmp/")
+
+			// Assertions
+			Expect(err).To(HaveOccurred())
+			Expect(result).To(BeNil())
+			Expect(err.Error()).To(ContainSubstring("Failed to open file /tmp/gpsegconfig_dump. Error: open /tmp/gpsegconfig_dump: no such file or directory"))
+		})
+
+		It("should return an error for a file with less than 9 number of fields", func() {
+			// Create a temporary file with incorrect fields content
+			content := "invalid_content\n"
+			tempConfFile := createSegConfigFile(content)
+
+			// Call the function under test
+			result, err := cluster.GetSegmentConfigurationFromFile(os.TempDir())
+
+			// Assertions
+			Expect(err).To(HaveOccurred())
+			Expect(result).To(BeNil())
+			Expect(err.Error()).To(ContainSubstring("Unexpected number of fields (1) in line: invalid_content"))
+
+			// Cleanup
+			os.Remove(tempConfFile.Name())
+		})
+
+		It("should return an error for a file with more than 10 number of fields", func() {
+			// Create a temporary file with incorrect fields content
+			content := "1 -1 p p n u 7000 localhost localhost /data/dir-1 dummy\n"
+			tempConfFile := createSegConfigFile(content)
+
+			// Call the function under test
+			result, err := cluster.GetSegmentConfigurationFromFile(os.TempDir())
+
+			// Assertions
+			Expect(err).To(HaveOccurred())
+			Expect(result).To(BeNil())
+			Expect(err.Error()).To(ContainSubstring("Unexpected number of fields (11) in line: 1 -1 p p n u 7000 localhost localhost /data/dir-1 dummy"))
+
+			// Cleanup
+			os.Remove(tempConfFile.Name())
+		})
+
+		It("should fail when there is type conversion error", func() {
+			// Create a temporary file with one invalid int field
+			content := "1a -1 p p n u 7000 localhost localhost /data/dir1\n"
+			tempConfFile := createSegConfigFile(content)
+
+			//Call the function under test
+			result, err := cluster.GetSegmentConfigurationFromFile(os.TempDir())
+
+			// Assertions
+			Expect(err).To(HaveOccurred())
+			Expect(result).To(BeNil())
+			Expect(err.Error()).To(ContainSubstring("Failed to convert dbID with value 1a to an int. Error: strconv.Atoi: parsing \"1a\": invalid syntax"))
+
+			//Cleanup
+			os.Remove(tempConfFile.Name())
+		})
+
+	})
+
 	Describe("GetSegmentConfiguration", func() {
 		header := []string{"dbid", "contentid", "role", "preferredrole", "mode", "status", "port", "hostname", "address", "datadir"}
 		localSegOneValue := cluster.SegConfig{1, 0, "p", "p", "s", "u", 6002, "localhost", "127.0.0.1", "/data/gpseg0"}
--- a/cluster/cluster_test.go
+++ b/cluster/cluster_test.go
@@ -76,10 +76,14 @@ var _ = Describe("cluster/cluster tests", func() {
 		})
 	})
 	Describe("GetSegmentConfiguration", func() {
-		header := []string{"contentid", "hostname", "datadir"}
-		localSegOne := []driver.Value{"0", "localhost", "/data/gpseg0"}
-		localSegTwo := []driver.Value{"1", "localhost", "/data/gpseg1"}
-		remoteSegOne := []driver.Value{"2", "remotehost", "/data/gpseg2"}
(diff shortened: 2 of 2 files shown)


cluster/cluster.go
```diff
@@ -563,3 +571,101 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn, getMirrors ...bool)
...
+		}
+		if coordinator_datadir == "" {
+			errStr := "COORDINATOR_DATA_DIRECTORY is not set, not able to generate gpsegconfig_dump file path"
+			gplog.Error(errStr)
+			return nil, fmt.Errorf(errStr)
+		}
+		conFilePath = filepath.Join(coordinator_datadir, filename)
+	} else {
+		conFilePath = file
+	}
```
Chibin, line 632: I'm hesitant on adding adding this (COORDINATOR_DATA_DIRECTORY and MASTER_DATA_DIRECTORY) check for gp-common-go-libs.
Which ever program who wants to use the environment mentioned above,  should have already checked it; the library should handle just the filepath.

That's just my take, I'll let someone else chime in.

kaknikhil: Yeah I agree the calling function should just pass the entire filename to this function

cluster/cluster.go
```diff
@@ -563,3 +571,101 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn, getMirrors ...bool)
...
+		1 -1 p p n u 7000 shrakeshSMD6M.vmware.com shrakeshSMD6M.vmware.com */
+		/* newer version of gpsegconfig_dump file has 10 parts as it does have datadir
+		1 -1 p p n u 7000 shrakeshSMD6M.vmware.com shrakeshSMD6M.vmware.com /data/qddir/demoDataDir-1 */
+		if parts != 9 && parts != 10 {
+			errStr := fmt.Sprintf("Unexpected number of fields (%d) in line: %s", parts, scanner.Text())
+			gplog.Error(errStr)
+			return nil, fmt.Errorf(errStr)
+		}
+
+		dbID, err := strconv.Atoi(fields[0])
```
kaknikhil, line 660: You can create a common function for the conversion
```
func convertField(fieldName string, fieldValue string) (int, error) {
	val, err := strconv.Atoi(fieldValue)
	if err != nil {
		errStr := fmt.Sprintf("Failed to convert %s with value %s to an int. Error: %s", fieldName, fieldValue, err.Error())
		gplog.Error(errStr)
		return -math.MaxInt, fmt.Errorf(errStr)
	}
	return val, nil
}
```

cluster/cluster.go
```diff
@@ -563,3 +571,101 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn, getMirrors ...bool)
...
+		seg := SegConfig{
+			DbID:      dbID,
+			ContentID: content,
+			Role:      fields[2],
+			PRole:     fields[3],
+			Status:    fields[4],
+			Mode:      fields[5],
+			Address:   fields[8],
+			Port:      port,
+			Hostname:  fields[7],
```
kaknikhil: We can move this to above the "Address" field to maintain the sequence

cluster/cluster.go
```diff
@@ -49,6 +53,10 @@ type SegConfig struct {
 	DbID      int
 	ContentID int
 	Role      string
+	PRole     string
```
kaknikhil: I'm not sure if introducing these properties will break backwards compatibility or cause any other dependency issues ?

kaknikhil: I would use the entire name here "PrimaryRole"

kaknikhil:
Few high level observations:

* Does this apply to gpdb6 as well ?
* How often is gpsegconfig_dump written ? Is it possible that the information in this file could be stale or incorrect ? Specially in failover scenarios ? 
* What is the recommended use of this function ? Should this only be called when the database is down ? 

* The commit message should 
** contain an example of the gpsegconfig_dump file so that it's easy to review the PR
** explain the need for introducing this function in gp-common-go-libs instead of the utility itself
** recommendations around when to call this function (this should also be in the function's docstring)


```
This MR introduces a function named GetSegmentConfigurationFromFile to parse the gpsegconfig_dump file so that any utility can get the configuration even if the database is down

<Reason for adding this to gp-common-go-libs>

More about gpsegconfig_dump:
gpsegconfig_dump file contains segment configuration info and it is created in the coordinator data dir via fts process. gpsegconfig_dump keeps updated segment configuration info and as this flat file can be used to get the configuration data even if the database is down. Here is an example
\```
1 -1 p p n u 7000 127.0.0.1 127.0.0.1
8 -1 m m s u 7001 127.0.0.1 127.0.0.1
2 0 p p s u 7002 127.0.0.1 127.0.0.1
5 0 m m s u 7005 127.0.0.1 127.0.0.1
4 2 p p s u 7004 127.0.0.1 127.0.0.1
7 2 m m s u 7007 127.0.0.1 127.0.0.1
3 1 p p s u 7003 127.0.0.1 127.0.0.1
6 1 m m s u 7006 127.0.0.1 127.0.0.1
\```

```

cluster/cluster.go
shrakesh: > You can create a common function for the conversion
> 
> ```
> func convertField(fieldName string, fieldValue string) (int, error) {
> 	val, err := strconv.Atoi(fieldValue)
> 	if err != nil {
> 		errStr := fmt.Sprintf("Failed to convert %s with value %s to an int. Error: %s", fieldName, fieldValue, err.Error())
> 		gplog.Error(errStr)
> 		return -math.MaxInt, fmt.Errorf(errStr)
> 	}
> 	return val, nil
> }
> ```

@kaknikhil I am confused about where should i keep the convertField function. 

1. As an anonymous function 
` convertField := func (fieldName string, fieldValue string) (int, error){`
 this way the scope of the function would be limited and specific to GetSegmentConfigurationFromFile()
 
 2. As a utility function:
    one option is that I need to write the function and test cases for AtoI, ItoA, and similar conversions to complete the utility which will increase the scope of the pr, and also in the common-go-libs currently, there is no specific package for conversion utility and I feel it doesn't fit as well in the gp libs as the caller can take care of conversion and related error message. 
    Another option is to keep it as the unexposed function in the cluster package. 
    
I feel keeping the function anonymous would be good to go, let me know your opinion.

cluster/cluster.go
shrakesh: Updated as "PreferredRole" to make it consistent with gp_segment_configuration output.

cluster/cluster.go
shrakesh: updated the sequence of all the fields to make it consistent with the gp_segment_configuration table output.

shrakesh:
> * Does this apply to gpdb6 as well ?

Yes, this will apply to gpdb6 as well, in gpdb6 gpsegconfig_dump file does not have datadir so the check is added in such a way that it will ignore the datadir for 6X. even if we backport the introduction of  datadir pr (https://github.com/greenplum-db/gpdb/pull/16922) , still it will work for older minor version of gpdb6. 

> * How often is gpsegconfig_dump written ? Is it possible that the information in this file could be stale or incorrect ? Specially in failover scenarios ?

The gpsegconfig_dump is in sync with gp_segment_configuration table, in the fts probe if there is any update in the segment configuration it will dump the same update to gpsegconfig_dump flat file.  so this is the most updated configuration data we have after database is down. 

> * What is the recommended use of this function? Should this only be called when the database is down ?

The main purpose of the API is to get the configuration when the database is down. there is already another API GetSegmentConfiguration() which can be used when the database is up.  In the API we have added an arg to give the path of the file which has similar structure as gpsegconfig_file.

> * The commit message should
>   ** contain an example of the gpsegconfig_dump file so that it's easy to review the PR
>   ** explain the need for introducing this function in gp-common-go-libs instead of the utility itself
>   ** recommendations around when to call this function (this should also be in the function's docstring)
> 

I have updated the pr description, I will update the commit message as well. Updated function doc string to provide more details about the function.

cluster/cluster.go
shrakesh: I doubt it will do, the lib will be used at compile time to create a binary. but for safer side I will try to notify about the pr on related slack channels.

cluster/cluster_test.go
```diff
@@ -75,6 +84,138 @@ var _ = Describe("cluster/cluster tests", func() {
 			Expect(cmd).To(Equal([]string{"ssh", "-o", "StrictHostKeyChecking=no", "testUser@some-host", "ls"}))
 		})
 	})
+
+	Describe("GetSegmentConfigurationFromFile", func() {
```
piyushc01, line 91: Add a test for multi-line config file containing data of multiple segments

cluster/cluster.go
kaknikhil: What can happen is that existing utilities calling the other function `GetSegmentConfiguration` will now get the new struct but the newly added properties won't be set since they are only set in `GetSegmentConfigurationFromFile`.  It's possible that a utility might try to access one of these properties and error out or worse have unintended side effects.

Is there a reason we decided to not initialize these properties while calling `GetSegmentConfiguration` as well?

cluster/cluster.go
```diff
@@ -563,3 +571,116 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn, getMirrors ...bool)
...
+	fd, err := os.Open(confFile)
+	if err != nil {
+		errStr := fmt.Sprintf("Failed to open file %s. Error: %s", confFile, err.Error())
+		gplog.Error(errStr)
+		return nil, fmt.Errorf(errStr)
+	}
+	defer fd.Close()
+
+	results := make([]SegConfig, 0)
+	scanner := bufio.NewScanner(fd)
```
kaknikhil, line 613: By default, a new scanner object has a 64KB as the max token size per line. This should be enough for us since our lines shouldn't really be that big but just something to be aware of

kaknikhil, line 645: By default, a new scanner object has a 64KB as the max token size per line. This should be enough for us since our lines shouldn't really be that big but just something to be aware of

cluster/cluster.go
```diff
@@ -563,3 +571,116 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn, getMirrors ...bool)
...
+			Address:       fields[8],
+			DataDir:       datadir,
+		}
+
+		results = append(results, seg)
+	}
+
+	/* validating error during file reading */
+	if err := scanner.Err(); err != nil {
+		errStr := fmt.Sprintf("error reading confFile %s: %s", confFile, err.Error())
```
kaknikhil: s/confFile/gpsegconfig_dump file

cluster/cluster.go
```diff
@@ -563,3 +571,116 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn, getMirrors ...bool)
...
+2 0 p p n u 6002 localhost localhost /data/temp2
+3 1 p p n u 6003 localhost localhost /data/temp3
+4 2 p p n u 6004 localhost localhost /data/temp4
+
+Recommended use of the api is to get populate the gpdbArray when database is down.
+if the database is up use GetSegmentConfiguration()/MustGetSegmentConfiguration()  API
+
+input:
+confFile - absolute path of the gpsegconfig_dump or similar structured file
+(e.g. /Users/shrakesh/workspace/gpdb/gpAux/gpdemo/datadirs/qddir/demoDataDir-1/gpsegconfig_dump)
```
kaknikhil: If the database is offline, how will the calling utility be able to get the coordinator data directory ?

cluster/cluster.go
```diff
@@ -563,3 +571,116 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn, getMirrors ...bool)
...
+
+return:
+Slice of populated SegConfig objects or else nil
+error if any or else nil.*/
+
+func GetSegmentConfigurationFromFile(confFile string) ([]SegConfig, error) {
```
kaknikhil: Won't the file always be named "gpsegconfig_dump" ? Why should the utility have to know this filename ? They can just pass the data directory and the function can append "gpsegconfig_dump" to the path
The only use case for asking the utility to pass the name is if this function can be called on arbitrary files that are not named gpsegconfig_dump

cluster/cluster.go
```diff
@@ -563,3 +571,116 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn, getMirrors ...bool)
...
+
+	/*Check if the given argument confFile is empty*/
+	if len(strings.TrimSpace(confFile)) == 0 {
+		errStr := fmt.Sprintf("Error: Given configuration file name is empty")
+		gplog.Error(errStr)
```
kaknikhil: Why are we logging all the errors with gplog ? Isn't that the utility's responsibility ?

cluster/cluster_test.go
```diff
@@ -75,6 +84,185 @@ var _ = Describe("cluster/cluster tests", func() {
...
+			Expect(err).To(BeNil())
+			Expect(result).To(HaveLen(1))
+			Expect(result[0]).To(Equal(expRes))
+
+			//Cleanup
+			tempConfFile.Close()
+			os.Remove(tempConfFile.Name())
+		})
+
+		It("should return expected result for a old (9 fields) gpsegconfig_dump file", func() {
```
kaknikhil: nit: s/a/an

cluster/cluster_test.go
```diff
@@ -75,6 +84,185 @@ var _ = Describe("cluster/cluster tests", func() {
...
+			tempConfFile := createSegConfigFile(content)
+
+			//call the function under test
+			result, err := cluster.GetSegmentConfigurationFromFile(tempConfFile.Name())
+			Expect(err).To(BeNil())
+			Expect(result).To(HaveLen(1))
+			Expect(result[0]).To(Equal(expRes))
+
+			//Cleanup
+			tempConfFile.Close()
```
kaknikhil: this should be in a defer

cluster/cluster.go
kaknikhil: Let me think about this a bit but think option 1 should be fine

cluster/cluster.go
```diff
@@ -563,3 +571,116 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn, getMirrors ...bool)
...
+gpsegconfig_dump file gets created at $COORDINATOR_DATA_DIR/gpsegconfig_dump by fts process
+The gpsegconfig_dump file follows a structured format, as illustrated in the example below:
+1 -1 p p n u 6000 localhost localhost /data/temp1
+2 0 p p n u 6002 localhost localhost /data/temp2
+3 1 p p n u 6003 localhost localhost /data/temp3
+4 2 p p n u 6004 localhost localhost /data/temp4
+
+Recommended use of the api is to get populate the gpdbArray when database is down.
+if the database is up use GetSegmentConfiguration()/MustGetSegmentConfiguration()  API
+
```
kaknikhil, line 615: We should also add a disclaimer here that the information returned by this function can be a bit stale since it depends on fts (user can configure fts to run less frequently)

cluster/cluster.go
shrakesh: the initial plan was to just fetch the $COORDINATOR_DATA_DIRECTORY/gpsegconfig_dump file and do the parsing on that, keeping it simple. but later we realized that since this is common go library for gp, we should provide an option to give the file path as an arg for arbitrary files.  this will also enable us to extend the API to parse different type of file format like (json) which we are using in project spine.  Yes we may have to add and enhance the logic but the signature will be same.

cluster/cluster.go
shrakesh: with context to https://github.com/greenplum-db/gp-common-go-libs/pull/85#discussion_r1448329077 as well.
COORDINATOR_DATA_DIRECTORY we can get it from the env variable.  with this API it is the caller's responsibility to provide a valid gpsegconfig_dump or similar file path. 

in the previous commit, we are appending the path inside the API to get the gpsegconfig_dump file path. but after the suggestion on PR I also agree that the providing file path as a user argument will keep the API use consistent. 

Another option we have is that we can have a supporting API to get the gpsegconfig_dump file path. 

please let me know your opinion.

cluster/cluster.go
shrakesh: yes, that's a good point thanks,  i will remove the logging.

cluster/cluster.go
shrakesh: IMHO if we are treating arg as a generic I think confFile name is ok, but if we change the solution to read only gpsegconfig_dump file then keeping the var name gpsegconfig_dump would make sense.

cluster/cluster.go
shrakesh: don't know what the purpose of not having all the SegConfig members the same as the gp_segment_configuration table. one thing I can think of the missing fields are ok to have nil value while using gpdbArray. 

your observation is valid that the user may accidentally access the new member assuming that it may have valid data but can find nil or empty. 

IMO 2 ways we can handle this issue 
1. keeping the struct the same as the previous and ignoring the missing member in GetSegmentConfigurationFromFile API 
2.  Update the GetSegmentConfiguration() API  to populate newly added members. 

Please provide your opinion, 2nd option looks more convincing to me.

kaknikhil: I think option 2 is right way to go but it would be nice if someone else also takes a look. What do you think @Chibin  ?
Also I'm not sure what option 1 means and how it solves the problem ?

cluster/cluster.go
kaknikhil: IMO the calling function shouldn't have to know about `gpsegconfig_dump` since they only care about getting the configuration details when the database is offline. So I think the calling utility can just pass the coordinator/master data directory.

Now that I'm thinking about it, maybe it's not so bad if the function doesn't accept anything and the function just figures out the master/coordinator env variable itself. This will make it really easy for the utility but it will mean that gp-common-go-libs will need to do the master/coordinator if check.  Need to think about this a bit more

cluster/cluster.go
kaknikhil: But this function will specifically be used for parsing the contents of gpsegconfig_dump. Unless we also want to use this function for a file with a different name but similar format/contents, the caller shouldn't really need to know about the existence of `gpsegconfig_dump`

cluster/cluster.go
Chibin: I think that... at the very least, we should do option 2.

----------------
Edit: playing around with the test cases

It looks like using `GetSegmentConfiguration` as is will not break. This is mainly because we specifically call out each column name, so right now, the new attributes added to the struct would be ignored.
So if someone where to assume that these are available... similar to `GetSegmentConfigurationFromFile` they will be empty.

If we never foresee someone using the new attributes when calling `GetSegmentConfiguration`, then perhaps it's okay to ignore?
But to play it safe, we should have them do the same thing where everything is populated.

cluster/cluster.go
Chibin: Are there any other type of file that can be read that's similar to `gpsegconfig_dump`?
I'm leaning towards `gpsegconfig_dump` as the name, but not a stopper.

cluster/cluster.go
Chibin: If this is being shown as the error to a user.
```
errStr := fmt.Sprintf("error reading confFile %s: %s", confFile, err.Error())
```
We should either put `gpsegconfig_dump file` or `configuration file` or `file` and not the variable name `confFile`.

kaknikhil:
Spoke to a few people about this and here is the summary
1. Is there any utility other than gpsupport that needs this function ? I think gpdr might also want to use it but need to think about that a bit more. The idea being that if there is only one utility for now, maybe we keep this function inside gpsupport. If there are at least two, then gp-common-go-libs might be fine
2. We have three options for the function signature
```
// Option 1
func GetSegmentConfigurationFromFile() {
	 if os.env("MASTER_DATA_DIRECTORY"):
	 	 dir := 
	 else os.env("COORDINATOR_DATA_DIRECTORY"):
	  	 dir :=

	filename := dir + "/gpsegconfig_dump"

	// read filename and return struct
}

// Option 2
func GetSegmentConfigurationFromFile(masterDataDir) {
	filename := masterDataDir + "/gpsegconfig_dump"
	// read filename and return struct
}

// Option 3
func GetSegmentConfigurationFromFile(filename) {
	// read filename and return struct

}

```

Option 2 seems to be the best out of these three because that way the utility doesn't need to know the name of the dump file and also gp-common-go-libs doesn't need to worry about the awkward if else

shrakesh:
> Spoke to a few people about this and here is the summary
> 
> 1. Is there any utility other than gpsupport that needs this function ? I think gpdr might also want to use it but need to think about that a bit more. The idea being that if there is only one utility for now, maybe we keep this function inside gpsupport. If there are at least two, then gp-common-go-libs might be fine
> 2. We have three options for the function signature
> 
> ```
> // Option 1
> func GetSegmentConfigurationFromFile() {
> 	 if os.env("MASTER_DATA_DIRECTORY"):
> 	 	 dir := 
> 	 else os.env("COORDINATOR_DATA_DIRECTORY"):
> 	  	 dir :=
> 
> 	filename := dir + "/gpsegconfig_dump"
> 
> 	// read filename and return struct
> }
> 
> // Option 2
> func GetSegmentConfigurationFromFile(masterDataDir) {
> 	filename := masterDataDir + "/gpsegconfig_dump"
> 	// read filename and return struct
> }
> 
> // Option 3
> func GetSegmentConfigurationFromFile(filename) {
> 	// read filename and return struct
> 
> }
> ```
> 
> Option 2 seems to be the best out of these three because that way the utility doesn't need to know the name of the dump file and also gp-common-go-libs doesn't need to worry about the awkward if else

on the first point: regarding keeping it in gp-common-go-libs, it was discussed with @ashwinstar that this API can be used in other utilities as well.  so keeping it in the common place would be preferred.  that is the reason we proceed with idea that any utility required for the configuration when the segment is down can be used and enhance the API accordingly.

on 2nd point: regarding passing arg as master_data_dir , this will be ok as this will have the option to parse arbitrary data dir which has gpsegconf_dump file.  that sounds reasonable. 
or we can combine option 1 and option 2 as well( via optional argument )  this also looks ok to me.

kaknikhil:
> on the first point: regarding keeping it in gp-common-go-libs, it was discussed with @ashwinstar that this API can be used in other utilities as well. so keeping it in the common place would be preferred. that is the reason we proceed with idea that any utility required for the configuration when the segment is down can be used and enhance the API accordingly.

👍 

> on 2nd point: regarding passing arg as master_data_dir , this will be ok as this will have the option to parse arbitrary data dir which has gpsegconf_dump file. that sounds reasonable. or we can combine option 1 and option 2 as well( via optional argument ) this also looks ok to me.

Let's go ahead with Option 2 then

cluster/cluster.go
shrakesh: > Are there any other type of file that can be read that's similar to `gpsegconfig_dump`? I'm leaning towards `gpsegconfig_dump` as the name, but not a stopper.

currently there is no other file similar to gpsegconfig_dump. as now we have decided to pass datadir as arg and appended gpsegconfig to create path, IMO now gpsegconfig_dump variable name look more sensible.

cluster/cluster.go
```diff
@@ -522,8 +531,12 @@ SELECT
 	s.dbid,
 	s.content as contentid,
 	s.role,
+	s.preferred_role,
```
Chibin: I don't think this works.

```suggestion
	s.preferred_role as preferredrole,
```

We should also update `GetSegmentConfiguration` test to reflect this change.

cluster/cluster_test.go
```diff
@@ -75,6 +87,179 @@ var _ = Describe("cluster/cluster tests", func() {
...
+			// Assertions
+			Expect(err).To(HaveOccurred())
+			Expect(result).To(BeNil())
+			Expect(err.Error()).To(ContainSubstring("Failed to convert dbID with value 1a to an int. Error: strconv.Atoi: parsing \"1a\": invalid syntax"))
+
+			//Cleanup
+			os.Remove(tempConfFile.Name())
+		})
+
+	})
```
Chibin, line 262: At the minimum, we should add the preferred role to this test in GetSegmentConfiguration; it should expose the issue I mentioned above.
```
It("returns only primaries for a single-host, single-segment cluster", func() {
			fakeResult := sqlmock.NewRows(header).AddRow(localSegOne...)
			mock.ExpectQuery("SELECT (.*)").WillReturnRows(fakeResult)
			results, err := cluster.GetSegmentConfiguration(connection)
			Expect(err).ToNot(HaveOccurred())
			Expect(len(results)).To(Equal(1))
			Expect(results[0].DataDir).To(Equal("/data/gpseg0"))
			Expect(results[0].Hostname).To(Equal("localhost"))
			Expect(results[0].PreferredRole).To(Equal("primary"))
})
```

cluster/cluster_test.go
```diff
@@ -75,20 +87,201 @@ var _ = Describe("cluster/cluster tests", func() {
...
 	Describe("GetSegmentConfiguration", func() {
-		header := []string{"contentid", "hostname", "datadir"}
-		localSegOne := []driver.Value{"0", "localhost", "/data/gpseg0"}
-		localSegTwo := []driver.Value{"1", "localhost", "/data/gpseg1"}
-		remoteSegOne := []driver.Value{"2", "remotehost", "/data/gpseg2"}
+		header := []string{"dbid","contentid","role","preferredrole", "mode", "status","port","hostname","address", "datadir"}
+		localSegOne := []driver.Value{1,0,"p","p","s","u", 6002,"localhost","127.0.0.1", "/data/gpseg0"}
+		localSegTwo := []driver.Value{2,1,"m","m","s","u", 6003,"localhost","127.0.0.1", "/data/gpseg1"}
+		remoteSegOne := []driver.Value{3,2,"p","m","s","u",6004, "remotehost","127.0.0.1", "/data/gpseg2"}
```
Chibin, line 273: This looks good to me, but before committing, can we run the `go fmt` for this file?

Chibin (approved):
LGTM!

I'd wait for @kaknikhil's approval as well

cluster/cluster_test.go
```diff
@@ -75,20 +87,201 @@ var _ = Describe("cluster/cluster tests", func() {
...
 
 		It("returns only primaries for a single-host, single-segment cluster", func() {
 			fakeResult := sqlmock.NewRows(header).AddRow(localSegOne...)
 			mock.ExpectQuery("SELECT (.*)").WillReturnRows(fakeResult)
 			results, err := cluster.GetSegmentConfiguration(connection)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(len(results)).To(Equal(1))
-			Expect(results[0].DataDir).To(Equal("/data/gpseg0"))
+			Expect(results[0].DbID).To(Equal(1))
+			Expect(results[0].ContentID).To(Equal(0))
```
kaknikhil: I think we should also add these new Expects to the other scenarios

cluster/cluster.go
```diff
@@ -563,3 +580,117 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn, getMirrors ...bool)
...
+		/* older version of gpsegconfig_dump has 9 parts as it doesn't have datadir
+			1 -1 p p n u 7000 shrakeshSMD6M.vmware.com shrakeshSMD6M.vmware.com
+		newer version of gpsegconfig_dump has 10 parts as it does have datadir
+			1 -1 p p n u 7000 shrakeshSMD6M.vmware.com shrakeshSMD6M.vmware.com /data/qddir/demoDataDir-1 */
+		if parts != 9 && parts != 10 {
+			errStr := fmt.Sprintf("Unexpected number of fields (%d) in line: %s", parts, scanner.Text())
+			return nil, fmt.Errorf(errStr)
+		}
+
+		dbID, err := convertField("dbID", fields[0])
```
kaknikhil: Now that we aren't logging the error using gplog, we might not need this function since it's a simple oneliner now

```
dbID, err := strconv.Atoi("dbID")
if err != nil {
	return nil, fmt.Sprintf("Failed to convert dbid with value %s to an int. Error: %s", ...)
}
```

cluster/cluster.go
```diff
@@ -563,3 +580,117 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn, getMirrors ...bool)
 	gplog.FatalOnError(err)
 	return segConfigs
 }
+
+/*GetSegmentConfigurationFromFile to parse gpsegconfig_dump file to retrieve segment configuration information.
```
kaknikhil: Suggested reword:
```

// GetSegmentConfigurationFromFile parses the gpsegconfig_dump file to retrieve segment configuration information.
// Recommended use of the api is to get the contents of gp_segment_configuration when database is down.
// If the database is up, use GetSegmentConfiguration()/MustGetSegmentConfiguration() instead.
//  
// gpsegconfig_dump file gets created at $COORDINATOR_DATA_DIR/gpsegconfig_dump by the fts process. The frequency of
// writing to this file is governed by various fts gucs. Note: Since the gpsegconfig_dump file is updated by fts process the // information returned by this function can be a bit stale since user can configure fts to run less frequently
// 
// The gpsegconfig_dump file follows a structured format, as illustrated in the example below:
// 1 -1 p p n u 6000 localhost localhost /data/temp1
// 2 0 p p n u 6002 localhost localhost /data/temp2
// 3 1 p p n u 6003 localhost localhost /data/temp3
// 4 2 p p n u 6004 localhost localhost /data/temp4
//
// <Include an example with data dir and mention the gpdb version for each of these examples> 

// Input: Coordinator data dir path
// It returns the populated SegConfig object and error if any
```

cluster/cluster.go
```diff
@@ -563,3 +580,117 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn, getMirrors ...bool)
...
+
+return:
+Slice of populated SegConfig objects or else nil
+error if any or else nil.*/
+
+func GetSegmentConfigurationFromFile(coordinatorDataDir string) ([]SegConfig, error) {
+
+	/*Check if the given argument coordinator_data_dir is empty*/
+	if len(strings.TrimSpace(coordinatorDataDir)) == 0 {
+		errStr := fmt.Sprintf("Error: Given coordinator data directory name is empty")
```
kaknikhil:
```
if len(strings.TrimSpace(coordinatorDataDir)) == 0 { 
    return nil, fmt.Errorf("Coordinator data directory path is empty")
}
```

cluster/cluster.go
```diff
@@ -563,3 +580,117 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn, getMirrors ...bool)
...
+	}
+
+	/*Generate gpsegconfig_dump file path*/
+	gpsegconfigDump := path.Join(coordinatorDataDir, "gpsegconfig_dump")
+
+	/* Open gpsegconfig_dump */
+	fd, err := os.Open(gpsegconfigDump)
+	if err != nil {
+		errStr := fmt.Sprintf("Failed to open file %s. Error: %s", gpsegconfigDump, err.Error())
+		return nil, fmt.Errorf(errStr)
```
kaknikhil: We can inline this return as well
```
return nil, fmt.Errorf("Failed to open file %s. Error: %w", gpsegconfigDump, err)
```

cluster/cluster.go
```diff
@@ -563,3 +580,117 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn, getMirrors ...bool)
...
+		parts := len(fields)
+		var datadir string
+
+		/* older version of gpsegconfig_dump has 9 parts as it doesn't have datadir
+			1 -1 p p n u 7000 shrakeshSMD6M.vmware.com shrakeshSMD6M.vmware.com
+		newer version of gpsegconfig_dump has 10 parts as it does have datadir
+			1 -1 p p n u 7000 shrakeshSMD6M.vmware.com shrakeshSMD6M.vmware.com /data/qddir/demoDataDir-1 */
+		if parts != 9 && parts != 10 {
+			errStr := fmt.Sprintf("Unexpected number of fields (%d) in line: %s", parts, scanner.Text())
+			return nil, fmt.Errorf(errStr)
```
kaknikhil: oneliner return here as well 
```
return nil, fmt.Errorf("Unexpected number of fields (%d) in line: %s", parts, scanner.Text())
```

cluster/cluster.go
```diff
@@ -563,3 +580,117 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn, getMirrors ...bool)
...
+			return nil, err
+		}
+
+		port, err := convertField("port", fields[6])
+		if err != nil {
+			return nil, err
+		}
+
+		// there are 10 fields in new version of gpsegconfig_dump file
+		if parts == 10 {
```
kaknikhil, line 677: I would create the datadir var here closer to it's usage rather than up at the top
```
datadir := ""
if parts == 10 {
	datadir = fields[9]
}
```

cluster/cluster.go
```diff
@@ -563,3 +580,117 @@ func MustGetSegmentConfiguration(connection *dbconn.DBConn, getMirrors ...bool)
...
+			DataDir:       datadir,
+		}
+
+		results = append(results, seg)
+	}
+
+	/* validating error during gpsegconfig_dump file read */
+	if err := scanner.Err(); err != nil {
+		errStr := fmt.Sprintf("Failed to read gpsegconfig_dump file %s: %s", gpsegconfigDump, err.Error())
+		return nil, fmt.Errorf(errStr)
```
kaknikhil: onliner return here as well
```
return nil, fmt.Errorf("Failed to read gpsegconfig_dump file %s: %w", gpsegconfigDump, err)
```

kaknikhil (approved):
LGTM. Make sure to rebase and squash all the commits and add all the context to the commit message as well.
It might make sense to separate out the commits for `GetSegmentConfigurationFromFile` and `GetSegmentConfiguration`

PR #87, opened 2024-04-17
Add new Progress command for gplog

khuddlefish:
This is to be used with progress bars. It uses the prefix INFO and logs at
verbose level to stdout and info level to the log file so it won't disrupt
progress bars but will give useful information in the log files.

### DIFF
--- a/gplog/gplog.go
+++ b/gplog/gplog.go
@@ -251,6 +251,25 @@ func Warn(s string, v ...interface{}) {
 	_ = logger.logStdout.Output(1, message)
 }
 
+/*
+ * Progress is for messages that show progress as an alternative to a progress bar.
+ * We write them to the log file if fileVerbosity is >= LOGINFO, and we write them to stdout if shellVerbosity >= LOGVERBOSE
+ */
+
+func Progress(s string, v ...interface{}) {
+	logMutex.Lock()
+	defer logMutex.Unlock()
+	var message string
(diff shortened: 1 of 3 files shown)


gplog/gplog.go
```diff
@@ -251,6 +251,25 @@ func Warn(s string, v ...interface{}) {
...
+/*
+ * Progress is for messages that show progress as an alternative to a progress bar.
+ * We write them to the log file if fileVerbosity is >= LOGINFO, and we write them to stdout if shellVerbosity >= LOGVERBOSE
+ */
+
+func Progress(s string, v ...interface{}) {
+	logMutex.Lock()
+	defer logMutex.Unlock()
+	if logger.fileVerbosity >= LOGINFO || logger.shellVerbosity >= LOGVERBOSE {
+		message := GetLogPrefix("INFO") + fmt.Sprintf(s, v...)
```
denalex: This method logs the same message with different levels to file and console respectively. While this is not consistent with the overall design, I see this as an optional utility that might be helpful in some cases. 

However, I think the `prefix` should match the logging level being used for the respective destinations. In other words, I think we should log with `INFO` prefix to file and `VERBOSE` prefix to console.

khuddlefish: That's valid. I considered that and would be open to doing it that way instead of INFO for both

Chibin (approved):
LGTM

kyeap-vmware (approved):

PR #88, opened 2024-04-22
Added separate shell prefix and colorization to gplog

denalex:
PXF uses `gplog` library to output to the console and log messages to the file, however we have the following requirements that are implemented by this PR. All existing functionality is intact, so the changes should be additive only.

1. Have separate prefixes for console and log file output. We do not want to have users see the "header" with timestamps, pids, etc as it pollutes the important message payloads. The prefixes are now separated and I included the default function that would only append logging level for WARNING, ERROR, and CRITICAL message levels.
2. We have colorized output to the console, so this functionality also has been added as a logger flag. When enbaled, the colorization will apply to ERROR and CRITICAL messages (red) , WARNING messages (yellow) and INFO messages logged with new `Success()` method only (green).

### DIFF
--- a/gplog/gplog.go
+++ b/gplog/gplog.go
@@ -105,8 +105,6 @@ type GpLogger struct {
 	shellVerbosity     int
 	fileVerbosity      int
 	header             string
-	shellHeader        string
-	fileHeader         string
 	logPrefixFunc      LogPrefixFunc
 	shellLogPrefixFunc LogPrefixFunc
 	colorize           bool
--- a/gplog/gplog.go
+++ b/gplog/gplog.go
@@ -215,6 +215,14 @@ func SetColorize(shouldColorize bool) {
 	logger.colorize = shouldColorize
 }
 
+// GetColorize returns whether the colorization of shell console output has been enabled
+func GetColorize() bool {
+	if logger == nil {
+		return false
+	}
+	return logger.colorize
+}
+
 func SetLogFileNameFunc(fileNameFunc func(string, string) string) {
 	logFileNameFunc = fileNameFunc
 }
--- a/gplog/gplog.go
+++ b/gplog/gplog.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"log"
 	"os"
+	"regexp"
 	"strings"
 	"sync"
 
@@ -59,6 +60,16 @@ const (
 	LOGDEBUG
 )
 
+// --- constants for colorized output support
+const escape = "\x1b"
+
+const (
+	NONE = iota
+	RED
+	GREEN
+	YELLOW
+)
+
 /*
  * Leveled logging output functions using the above log levels are implemented
  * below.  Info(), Verbose(), and Debug() print messages when the log level is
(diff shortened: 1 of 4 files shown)


gplog/gplog.go
```diff
@@ -59,6 +60,16 @@ const (
 	LOGDEBUG
 )
 
+// --- constants for colorized output support
+const escape = "\x1b"
+
+const (
```
tatb-vmware, line 69: Go's enum type is a little odd compared to other languages. The "idiomatic" method to go about it is to have a `type Color int` before the const `iota` block, such that any function that takes in these `ints` can be less confusing by taking in the custom `Color` type instead of the primitive `int` type.

An example here: https://go.dev/doc/effective_go#initialization

gplog/gplog.go
```diff
@@ -59,6 +60,16 @@ const (
 	LOGDEBUG
 )
 
+// --- constants for colorized output support
+const escape = "\x1b"
```
tatb-vmware:
```suggestion
const ESCAPE = "\x1b"
```
Can we capitalize it such that we can denote it's a const?

tatb-vmware (approved):

gplog/gplog.go
```diff
@@ -190,13 +228,36 @@ func defaultLogPrefixFunc(level string) string {
...
 	if logger.logPrefixFunc != nil {
 		return logger.logPrefixFunc(level)
 	}
 	return defaultLogPrefixFunc(level)
 }
 
+// GetShellLogPrefix returns a prefix to prepend to the message before sending it to the shell console
+// Use this mechanism if it is desired to have different prefixes for messages sent to the console and to the log file.
+// A caller must first set a custom function that will provide a custom prefix by calling SetShellLogPrefixFunc.
+// If the custom function has not been provided, this function returns a prefix produced by the GelLogPrefix function,
```
axue-broadcom:
```suggestion
// If the custom function has not been provided, this function returns a prefix produced by the GetLogPrefix function,
```

axue-broadcom (approved):

gplog/gplog.go
```diff
@@ -87,15 +101,21 @@ type LogFileNameFunc func(string, string) string
...
+	fileVerbosity      int
+	header             string
+	logPrefixFunc      LogPrefixFunc
+	shellLogPrefixFunc LogPrefixFunc
+	colorize           bool
+}
+
+// levelsToPrefix is a regex for determining if the message level will be shown on console
+// by the DefaultShortLogPrefixFunc function
+var levelsToPrefix = regexp.MustCompile(`WARNING|ERROR|CRITICAL`)
```
khuddlefish: Can this be declared closer to where it is used?

gplog/gplog.go
```diff
@@ -265,63 +347,66 @@ func Progress(s string, v ...interface{}) {
...
+		suffix += fmt.Sprintf("%v", err)
 		stackTraceStr = formatStackTrace(errors.WithStack(err))
 		if s != "" {
-			message += ": "
+			suffix += ": "
 		}
 	}
-	message += strings.TrimSpace(fmt.Sprintf(s, v...))
+	suffix += strings.TrimSpace(fmt.Sprintf(s, v...))
+	message := GetLogPrefix("CRITICAL") + suffix
```
khuddlefish: The variable names feel a little strange to me here building up most things as `suffix`. I think of suffix as being after the main content, not containing it. I think it might feel more clear to me to build up the content as `message` and then create a new variable `logMessage` where you prepend the LogPrefix and `shellMessage` where you prepend the shell prefix? Or technically you could append them together when you do output like 

`_ = logger.logFile.Output(1, GetLogPrefix("CRITICAL")+message+stackTraceStr)` and
`_ = logger.logFile.Output(1, GetShellLogPrefix("CRITICAL")+message+stackTraceStr)`

denalex: renamed `suffix` -> `message` and `message` -> `fullMessage`

PR #89, opened 2024-05-06
Add `SelectContext` and `QueryContext` to DBConn

jnihal:
These new methods allow us to pass a context when executing queries. This can be helpful in situations where we want to cancel a running query by canceling the context passed.


Annu149 (approved):
LGTM

jimmyyih (approved):

PR #90, opened 2024-05-17
Create Custom log command and use it for CheckClusterError

khuddlefish:
Replace Progress command with Custom command for more logging control.

The Progress command was just for doing info level to the logfile and verbose
level to stdout for table progress in gpbackup and gprestore. However, I found
at least one more place where using different settings for terminal and logfile
would be useful. We CheckClusterError in the cluster package for some commands
and assume we will log details to the file, but not to stdout.

Since this is a more general use case, we can make a more general command that
lets the caller pass in values for file threshold and shell threshold
separately to accommodate more use cases.

Then we can use the new command for CheckClusterError to maintain the usual behavior
of always writing errors to the log file but only writing to the shell if the level is verbose or higher.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -388,10 +388,10 @@ func (cluster *Cluster) CheckClusterError(remoteOutput *RemoteOutput, finalErrMs
 		case func(content int) string:
 			content := failedCommand.Content
 			host := cluster.GetHostForContent(content)
-			gplog.Verbose("%s on segment %d on host %s %s", getMessage(content), content, host, errStr)
+			gplog.Custom(gplog.LOGERROR, gplog.LOGVERBOSE, "%s on segment %d on host %s %s", getMessage(content), content, host, errStr)
 		case func(host string) string:
 			host := failedCommand.Host
-			gplog.Verbose("%s on host %s %s", getMessage(host), host, errStr)
+			gplog.Custom(gplog.LOGERROR, gplog.LOGVERBOSE, "%s on host %s %s", getMessage(host), host, errStr)
 		}
 		gplog.Verbose("Command was: %s", failedCommand.CommandString)
 	}
--- a/gplog/gplog.go
+++ b/gplog/gplog.go
@@ -294,6 +294,20 @@ func SetErrorCode(code int) {
 	errorCode = code
 }
 
+func getVerbosityString(verbosity int) string {
+	switch verbosity {
+	case LOGERROR:
+		return "ERROR"
+	case LOGINFO:
+		return "INFO"
+	case LOGVERBOSE:
+		return "DEBUG"
+	case LOGDEBUG:
+		return "DEBUG"
+	}
+	return ""
+}
+
 /*
  * Log output functions, as described above
  */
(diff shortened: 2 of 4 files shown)


cluster/cluster.go
```diff
@@ -388,10 +388,10 @@ func (cluster *Cluster) CheckClusterError(remoteOutput *RemoteOutput, finalErrMs
 		case func(content int) string:
 			content := failedCommand.Content
 			host := cluster.GetHostForContent(content)
-			gplog.Verbose("%s on segment %d on host %s %s", getMessage(content), content, host, errStr)
+			gplog.Custom(gplog.LOGERROR, gplog.LOGVERBOSE, "%s on segment %d on host %s %s", getMessage(content), content, host, errStr)
```
denalex, line 391: I find this very unusual (at least from the perspective of Java's Log4J logging standard) that a single message gets logged and displayed with different logging levels (called `verbosity` here). IMO, the purpose of the log is to capture what is shown to the user (and maybe more context, if needed), so that during troubleshooting it would be possible to see what the user saw and more context around it to assist in troubleshooting. If this is not presented as an "ERROR" to the shell, why would it present as an "ERROR" to the log file ?

I understand the desire to limit the amount of info in the log file, but that is usually achieved by setting different logging level thresholds (called `verbosity` here) for file appender vs console appender (called `shell` here).

Having said that, PXF is using neither this particular function nor the Custom() function, so this does not seem to have an impact on the PXF project.

gplog/gplog.go
```diff
@@ -414,6 +409,36 @@ func Fatal(err error, s string, v ...interface{}) {
...
+
+func Custom(customFileVerbosity int, customShellVerbosity int, s string, v ...interface{}) {
+	logMutex.Lock()
+	defer logMutex.Unlock()
+	var message string
+	if logger.fileVerbosity >= customFileVerbosity {
+		message = GetLogPrefix(getVerbosityString(customFileVerbosity)) + fmt.Sprintf(s, v...)
+		_ = logger.logFile.Output(1, message)
+	}
+	if customShellVerbosity == LOGERROR {
```
denalex, line 424: does this need to be treated separately? I believe `ERROR` has the code of `0` so if only the `else` condition is used `logger.shellVerbosity >= customShellVerbosity`, then a custom message with ERROR level will always be `<=` to whatever shellVerbosity is set to. In other words, an ERROR `customShellVerbosity` will always be logged no matter what `logger.shellVerbosity` is set to since `logger.shellVerbosity >=0` as long as `logger.shellVerbosity` is not negative.

If this is only done for colorization of ERROR, then I'd understand. However, this method cannot be used for WARNINGs, and the colorization of WARN will not happen properly, as there is no concept of WARN level (verbosity), only a helper function `Warn` that does some custom stuff without respect to logging levels and thresholds.

bmdoil (commented):
> The Progress command was just for doing info level to the logfile and verbose
level to stdout for table progress in gpbackup and gprestore.

This feels like room for improvement in the progress bar implementation we use. It should probably be handling formatting based on the output location of file/shell. When we circle back to implementing parallel progress bars would be a good time to investigate that.

I think having a Custom command in general is fine. Maybe we let gpbackup/gprestore extend Custom by implementing Progress in utils/progress_bar.go instead of here, especially if gpbackup is the only project using it?

cluster/cluster.go
khuddlefish: The reason we did it this way is because it checks the errors on all segments (potentially hundreds or thousands) and it could be an overwhelming amount of log messages to output to the shell. On the shell side we summarize it to let the user know there were errors and how many and then we direct them to the main log file to see details.

gplog/gplog.go
khuddlefish: I separated it out because ERROR is the only one we output to `stderr` and everything else gets output to `stdout`. I put the colorization in for error since it was easy to do with it separated out. I did notice we don't have coverage for `Warn` with this since it's not actually a log level, but I didn't see a good way to cover that behavior.

bmdoil (approved):
LGTM

PR #91, opened 2024-05-21
Add length check on segmentList to avoid panic

tatb-vmware:
This PR adds a length check on the `segmentList` within `getSegmentByRole()` before returning the first index to avoid panicking.


khuddlefish (approved):
LGTM

PR #94, opened 2024-08-02
Retry failed ssh commands in ExecuteClusterCommand

khuddlefish:
We pass in maxAttempts and retry until the command passes. If there is an error, we collect it in the command.RetryError variable. We have ways to log messages for commands that were retried but eventually passed, along with continuing to log failed commands like we used to.

We test the command retry with a script that increments a number and exits 1 until the number gets high enough to simulate failing and then passing..


cluster/cluster.go
```diff
@@ -386,6 +425,19 @@ func (cluster *Cluster) GenerateAndExecuteCommand(verboseMsg string, scope Scope
 }
 
 func (cluster *Cluster) CheckClusterError(remoteOutput *RemoteOutput, finalErrMsg string, messageFunc interface{}, noFatal ...bool) {
+	for _, retriedCommand := range remoteOutput.RetriedCommands {
+		switch messageFunc.(type) {
+		case func(content int) string:
+			content := retriedCommand.Content
+			host := cluster.GetHostForContent(content)
+			gplog.Debug("Command failed before passing on segment %d on host %s with error:\n%v", content, host, retriedCommand.RetryError)
```
kyeap-vmware, line 432: The debug line here may not always be right. What happens if command is retried and all retries fail in the end?

cluster/cluster_test.go
```diff
@@ -556,58 +556,202 @@ var _ = Describe("cluster/cluster tests", func() {
...
+			scriptFile, _ := os.OpenFile(path.Join(testDir, "incr.bash"), os.O_RDWR|os.O_CREATE|os.O_EXCL, 0777)
+			// This script increments a number in a file and returns an error until it reaches 3 and returns success
+			scriptFmt := `#!/bin/bash
+n=$(cat %s/num.txt)
+m=$((n+1))
+if [ "$m" -lt "3" ]; then
+	echo $m > %s/num.txt
+	exit 1
+fi`
+			scriptFile.WriteString(fmt.Sprintf(scriptFmt, testDir, testDir))
```
kyeap-vmware, line 578: I'm getting an error in Ubuntu. `Text file busy`. Looks like Ubuntu wants scriptFile to be closed before it can be used by the test.
```
scriptFile.WriteString(fmt.Sprintf(scriptFmt, testDir, testDir))
scriptFile.Close()
```

cluster/cluster_test.go
```diff
@@ -556,58 +556,202 @@ var _ = Describe("cluster/cluster tests", func() {
...
+					Error:         fmt.Errorf("command error"),
+				}
+				remoteOutput = &cluster.RemoteOutput{
+					Scope:          0,
+					NumErrors:      1,
+					Commands:       []cluster.ShellCommand{failedCmd},
+					FailedCommands: []cluster.ShellCommand{failedCmd},
+				}
+			})
+			DescribeTable("CheckClusterError", func(scope cluster.Scope, includeCoordinator bool, perSegment bool, remote bool) {
```
kyeap-vmware: `includeCoordinator` looks unused in this test. We can remove?

kyeap-vmware (changes requested):

cluster/cluster.go
```diff
@@ -337,23 +344,55 @@ func (executor *GPDBExecutor) ExecuteLocalCommandWithContext(commandStr string,
...
 	length := len(commandList)
 	finished := make(chan int)
 	numErrors := 0
 	for i := range commandList {
 		go func(index int) {
+			var (
+				out    []byte
+				err    error
+				stderr bytes.Buffer
+			)
```
tatb-vmware, line 378: I'm curious why the `out` and `err` declaration were pulled out of line 385.

cluster/cluster.go
```diff
@@ -9,13 +9,15 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	joinerrs "errors"
```
tatb-vmware, line 12: I'm curious why the "errors" package was renamed from it's default name.

cluster/cluster.go
```diff
@@ -337,23 +344,55 @@ func (executor *GPDBExecutor) ExecuteLocalCommandWithContext(commandStr string,
...
+				err    error
+				stderr bytes.Buffer
+			)
 			command := commandList[index]
-			var stderr bytes.Buffer
-			cmd := command.Command
-			cmd.Stderr = &stderr
-			out, err := cmd.Output()
+			for attempt := 1; attempt <= maxAttempts; attempt++ {
+				stderr.Reset()
```
tatb-vmware, line 381: Do we also want to flush stdout if there were multiple attempts?

cluster/cluster.go
```diff
@@ -337,23 +344,55 @@ func (executor *GPDBExecutor) ExecuteLocalCommandWithContext(commandStr string,
...
+					newRetryErr := fmt.Errorf("attempt %d: error was %w: %s", attempt, err, stderr.String())
+					command.RetryError = joinerrs.Join(command.RetryError, newRetryErr)
+					if attempt != maxAttempts {
+						time.Sleep(retrySleep)
+					}
+				}
+			}
 			command.Stdout = string(out)
 			command.Stderr = stderr.String()
 			command.Error = err
```
tatb-vmware, line 397: I want to double check my understanding of the new retry behavior and how it affects existing the `command.Error`.

Suppose there are two attempts for running a particular command. Attempt 1 fails so we retry on attempt 2. Attempt 2 succeeds so now `Command.Error` is `nil` but `command.RetryError` has the first error inside. If this is correct, should the number of errors, `numErrors`, be 0 or 1? I wondering if we should pass that info in the return of `NewRemoteOutput`.

cluster/cluster.go
khuddlefish: We have another package called `errors` that we already use "github.com/pkg/errors". I didn't want to risk switching out the existing uses of that package for the new one and they would have a name conflict otherwise so I just brought in the new one under an alias for the specific use case.

cluster/cluster.go
khuddlefish: The code was moved into a for loop so if it had been kept where it was, they wouldn't have been accessible below outside of the for loop.

cluster/cluster.go
khuddlefish: The stdout gets refreshed when we create the new command with resetCmd

cluster/cluster.go
khuddlefish: I wanted numErrors to only reflect the number of commands that were totally failed at the end

cluster/cluster.go
khuddlefish: If a command fails until the end of maxAttempts it will end up in the failedCommands list rather than the retriedCommands list. Commands will only end up in one of those two lists. So any command in this list was retried but eventually passed.

cluster/cluster_test.go
khuddlefish: Interesting. I see we passed different values below so I think it was intended to be used. Maybe I dropped part of the test by mistake or something.

tatb-vmware (approved):

cluster/cluster.go
tatb-vmware: On the off chance we use more things from the stdlib `errors` library, maybe we can rename `joinerr` to `stderr` or something similar to reflect it's the Go standard error lib.

cluster/cluster_test.go
khuddlefish: I checked and it seems like all the information the functions needs about that is included in scope. I removed the test variable

khuddlefish:
@kyeap-vmware I think I addressed your comments. Anything else?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/greenplum-db/gp-common-go-libs
 
-go 1.19
+go 1.21
```
Chibin, line 3: Since the patch version is not being used here, does it pick up the latest patch version available on the machine?

khuddlefish: I think this is just a minimum version requirement. The 1.19 version was working on my machine even though I had 1.21 actually installed. I bumped it because I believe the new errors.Join functionality I started using was introduced in go 1.20 or 1.21. We already use go 1.21 in some other projects so I figured we could just match that version.

kyeap-vmware (approved):
LGTM!

